PR #4, opened 2024-05-04
Adding members for storing and retrieving Qubit properties

### LINKED ISSUE #3: Required struct members for storing Qubit properties
Currently, there are no struct members to store qubit properties (like decoherence times - T1 & T2, readout error, etc.)
https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/ee8495ee56fed1c5d26b6197ca58da3e5e158631/include/qdmi_internal.h#L137-L142

T1, T2, readout error, and readout length could be essential qubit properties for all backends (at least superconducting ones). 

I suggest adding members to store these properties. I'll also add a reference implementation of querying these properties in the Qiskit backend.

If the backend doesn't provide these properties, these values may be set to pre-defined error values (maybe some negative values).

Durganshu:
This PR addresses the issue #3 

Prominent changes are:

1. `QDMI_Qubit` now has members for storing T1, T2, Readout Length, Readout Error
2. New query functions for checking whether a qubit property exists.
3. A reference implementation for querying properties on the IBM backend


burgholzer (commented):
Hey 👋🏻 
Just quickly looked over the code changes and they mostly look good to me. 👍🏻
I added a couple of comments with some questions that might be worth exploring.

kayaercument (approved):

Issue #26, opened 2024-06-11
Rewriting the test


kayaercument:
The `openclose.c` uses a lot of implementation details, it needs to be re-written.

PR #27, opened 2024-06-11
Internal header moved to src

### LINKED ISSUE #15: Should `qdmi_internal.h` be internal?
The name *internal* conveys that it should not be exposed to neither other tools using QDMI nor backends implementing the QDMI. I see two options here
1. rename or merge it with one of the two other headers, or
2. refactor the code such that this header is actually only used in QDMI internally.

kayaercument:
`qdmi_internal.h` is moved to src. In this way, we can hide the implementation details. However, the test file `openclose.c` uses a lot implementation details. The test will be re-written (Please see Issue #26 ).

FYI @ystade


burgholzer (changes requested):
I am not sure whether I fully agree with this solution.
I get the idea of hiding implementation details (which is important). At the same time, header files should not be going into a `src` folder in my opinion. They belong in the `include` folder.

One alternative solution to the one proposed here would be to create an `include/detail` subfolder and place all internal headers there. That subfolder would then also not be part of the `PUBLIC_QDMI_HEADERS` property and it can be clearly communicated that these headers are for internal use only.
`pybind11` for example uses a `detail` subfolder that contains all the implementation details (https://github.com/pybind/pybind11/tree/stable/include/pybind11).

@ystade could you whey in on this as well?

burgholzer (approved):
LGTM ✅ many thanks!

Would be nice to get the tests fixed asap. It's never really ideal to have commented code lying around.

PR #29, opened 2024-06-17
implementing QDMI_core_device_count and QDMI_core_open_device

kayaercument:
To query the devices, I implement QDMI_core_device_count and QDMI_core_open_device.

Even though there is QDMI_Session in both function signatures, I did not use them. Do we need to remove them, or do we need to change the implementation?

FYI @echavarria-lrz


burgholzer (changes requested):
Just quickly looked through the changes here and left some comments.

burgholzer (approved):
I just have two minor suggestion that I'll quickly apply after the review.
Otherwise this LGTM.

I'll spend some time over the coming days on QDMI itself and might touch these implementation in the process. So I think it is best to get this in so that there won't be any conflicts.

PR #33, opened 2024-06-28
Implementing device property query for char

### LINKED ISSUE #23: Identification of the available devices
Selectors and schedulers might use device-based models to optimize their performance. Therefore, both of them require a kind of identification. We can use `BACKEND_NAME` or create another attribute named `BACKEND_ID`, which would be an integer.

We need to consider the production phase as well. How can we define the IDs? They might be environment variables, and we can assign them to the .qdmi-config file.

kayaercument:
I implemented QDMI_query_device_property_c to get backend information as a string. QDMI_Device_property was defined as a struct, and I changed it back to int.  

FYI @echavarria-lrz @Durganshu

### DIFF
--- a/include/qdmi.h
+++ b/include/qdmi.h
@@ -114,7 +114,6 @@ typedef int QDMI_Gate_property;
 // typedef int QDMI_Qubit_property;
 
 typedef int QDMI_Device_property;
-//typedef struct QDMI_Device_property_impl_d *QDMI_Device_property;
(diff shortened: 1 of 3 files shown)


burgholzer (changes requested):

burgholzer (approved):
LGTM. Thanks for quickly adapting to the feedback 😌

PR #38, opened 2024-07-16
assigning QDMI_query_device_property_c

### LINKED ISSUE #37: QDMI_query_device_property_c cannot be used
After the device is opened, we cannot call the QDMI_query_device_property_c. It raises a null exceptions.

kayaercument:
Just one liner


burgholzer (approved):
LGTM 👍🏻

PR #57, opened 2024-09-20
NULL Initialization of qdmi_library_list

### LINKED ISSUE #56: NULL Initialization of qdmi_library_list
In the `QDMI_session_init` function, we allocate memory for `QDMI_Session`. Since `qdmi_library_list` is a part of the session, memory is also allocated for it. From time to time, `qdmi_library_list` is not `NULL`, and it creates a problem while loading the libraries. To be sure, it needs to be `NULL` Initialized.

kayaercument:
Please see #56 .


burgholzer (changes requested):

burgholzer (approved):
LGTM 👍🏻

PR #61, opened 2024-10-08
🎨 Refine and Unify Interface Structure

### LINKED ISSUE #8: Remove packing routines from the device's side
Why do we need to pack the circuit on the device's side?

The idea is to let the vendors modify the QIR as needed? QLM's appliance, for example, always introduces measurements at the end. Should their QDMI backend remove the measurement from the QIR? I cannot think of a better reason for this. But I think we should always pack on our side to reduce latency.

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/4d4d0dc7c8331f0b2dc43119f2da0a53d2a5c203/include/qdmi_internal.h#L53

### LINKED ISSUE #16: Different Style of Querying a Property (Two ways to query num. qubits)
This issue is related to other issues and PRs (#2, #3, #4, #6). For clarity, I open this new issue.

At the moment there are in principle two ways to query the number of qubits, i.e., via

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/4d4d0dc7c8331f0b2dc43119f2da0a53d2a5c203/include/qdmi.h#L210

or

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/4d4d0dc7c8331f0b2dc43119f2da0a53d2a5c203/include/qdmi.h#L200

with 

```c
int num_qubits;
int QDMI_query_device_property_i(dev, NUM_QUBITS, &num_qubits);
```

However, for example the WMI backend only implements the first option. Besides the fact, that there should in all cases only be exact one way to query some property, I think those two methods also illustrate two different styles to query properties. I want the latter to be the center of this issue and discuss, the pros and cons of both approaches:

1. The first option introduces for every property its own function. This approach has the disadvantage that it is difficult to extend the interface since you have to define a new function for each property. On the other side, you have the advantage, that you can have custom parameters with respect to the property, i.e., a struct where you can represent a coupling map in a reasonable way.

2. For the second option you just need to add a new constant for new parameters and since those are just integers those new constants are not breaking interfaces from an older version, they will just not support the new properties. However, with this option I have my difficulties to think of a good solution to retrieve the coupling map for example since one would only have standard query function for the default datatype `int`, `float`, `double`, ...

To summarize, I think we should settle on one option and use it consistently and in particular prevent the cases where one property can be retrieved in two different ways as it is possible right now for the number of qubits.

### LINKED ISSUE #19: `QDMI_control_readout_raw_num` scales poorly w.r.t. the number of qubits
Backends return the measurement counts in a $2^{|\text{qubits}|}$-bit array:

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/3b86d11fa19f6aa2cb13502a6278593289351e26/src/qdmi_backend_q7.c#L82-L83

How should we address this exponential growth?

### LINKED ISSUE #20: New routine to query the supported QIR profile
Do we need to create a new routine to ask a target device what is the QIR profile they support?

https://github.com/qir-alliance/qir-spec/tree/93597a337816e260628eb4d53b8510427251d4a9/specification/under_development/profiles

Or should we use a `QDMI_query_device_property_` routine?

### LINKED ISSUE #21: Adding execution time to `QDMI_Gate_impl_d`
@AdVetter @martin-knudsen @alfrisch @Durganshu Should we also add execution time as a property for each gate?

@flowerthrower @Gistbatch Would this benefit our schedulers?

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/3b86d11fa19f6aa2cb13502a6278593289351e26/include/qdmi_internal.h#L127-L135

### LINKED ISSUE #26: Rewriting the test

### LINKED ISSUE #34: Renaming QDMI_Device_property
`QDMI_Device_property` is defined as an int and used in device query function. As far as I see, `QDMI_Device_property` is meant to be used as an index. @burgholzer suggest we need to rename the variable accordingly.

FYI @mnfarooqi @echavarria-lrz

### LINKED ISSUE #40: Renaming `QDMI_core_device_count` to `QDMI_core_library_count`
https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/b966745af7187994496ec99d89b4426cea62e96a/src/qdmi_core.c#L584-L597

I would suggest renaming this routine to `QDMI_core_library_count` as it doesn't reflect the number of actual devices.

### LINKED ISSUE #43: Multiple QDMI_Session can't remain simultaneously active
While testing the integration of sys-sage and FoMaC, we observed that if one `QDMI_Session` is already active, opening another `QDMI_Session` won't work. This issue can be further investigated.

### LINKED ISSUE #54: Packing the fragment size
The vendors are reporting to us that they need the size of the IR. They can work around it if QASM is packed. However, they would need the size of the QIR code. Even though we have `sizebuffer` in the QDMI_Frag, we don't have a function to pack it. 

FYI @echavarria-lrz

ystade:
This PR targets a couple of existing issues with the current implementation of the interface. It refactors and unifies the existing functionalities and organizes them in a clear and concise structure.

Key features of the refactored interface:
- Devices are managed by the session
- Distinct headers for the backend to implement and the interface itself
- Merging the library struct into the device
- Unified and clean way of querying different properties
- Clustering properties in three blocks:
  - Device properties
  - Site properties
  - Operation properties
- Capability to extend the interface without breaking compatibility
- Initial documentation setup based on Doxygen and sphinx
  
Fixes:
- #54 as there is no `QDMI_Fragment` anymore because the QASM or QIR blob is submitted directly to the backend
- #43
- #40 as the device carries all function pointers itself, there is no library anymore
- #34 fixed by clear nomenclature and extensive documentation 
- #26 
- #21 in particular, the interface is now easier to extend without breaking the compatibility
- #20 can easily be added as another property if needed
- #19 as documented in the respective function
- #16 
- #8
- #7 as there is a cleaned header for the backend
- #6


schulzm (commented):
I have some concerns on the memory management of returned values, as well as on the interactions when opening/closing devices that should go over.

schulzm (commented):
Made it though most of it, will try the rest later, but need to head to plane first, and hence committing the first set.

PR #65, opened 2024-10-25
✨ Example for QDMI Client

ystade:
### Issue
Right now, there is an example implementation for a QDMI backend and a QDMI driver. However, the third component using the QDMI client interface is missing.

### Solution
This PR implements a small C++ library that provides some easy-to-use functions as a wrapper around QDMI functions. Additionally, it shows a simple compilation pass using this library.


ystade:
Esp. @burgholzer but also to the rest. Should we already call the library fomac since this will never be the proper fomac. Additionally, I was wondering whether the library should be implemented as more object-oriented, i.e., the Site and the Operation could be a class, and properties about them could be queried via member functions.

burgholzer (approved):
Many thanks for adding this 🙏🏼 

I just quickly went through the code and adapted it slightly so that we can get it in. Hopefully, the individual commits should rather clearly show the respective changes.
Most importantly:
- I prefixed a couple library targets, headers and source files with `example_` to more clearly differentiate them from real QDMI library code.
- I fixed a compilation error on linux regarding the driver. The compiler complained that position independent code shall be used.
- I fixed a runtime error because of a double free on linux which was caused by freeing the results in a job when they have not been allocated.

As for the more object-oriented FoMaC implementation: I am all for it, but would rather defer that to a separate PR. I'd prefer to iterate quickly here for now in comparison to getting ever-growing PRs. Maybe turn this into an issue?

Next thing tomorrow is to set up proper CI so that we can avoid merging code that does not compile under the systems we aim to support and that code formatting is automatically ensured.
I'll take that on tomorrow. 

Until then, let's roll with this 🚀

PR #71, opened 2024-10-28
✨ Add C++ Example Device

ystade:
### Issue

The existing example device is purely implemented in C. Since it is also possible to implement a C interface such as QDMI in C++, we also want to provide an example of implementation in C++. Programmers might be more familiar with C++ than C, but at least they can now choose between those two examples.

### Solution

This PR adds an example device implemented in C++, returning the same data as the C device (as of now).

### Remarks
This PR additionally
- fixes some issues in the docstrings of the interface and
- extends the tests to properly test job management.


burgholzer (approved):
Thanks for preparing this @ystade! 🙏🏼 

I iterated the PR slightly so that we can get it in.
Hopefully, the individual commits help to understand the changes.

Notably, I removed a level of indirection in the C++ device by not dynamically allocating the Sites and Operations and moving all the global data into an anonymous namespace (as suggested by clang-tidy).
I also adjusted the PR title and description to use "device" instead of "backend".

Let's get this in and see whether we can get the CI PR running.

PR #72, opened 2024-10-28
✨ Add Project Templates for Devices

ystade:
### Issue

When someone wants to implement a device from scratch, they would need to figure out their project structure by themselves.

### Solution

This PR provides a template for implementing a device in C or C++. Within the template, CMake is set up accordingly and links basic tests.


burgholzer (approved):
Also here: many thanks for setting this up 🎉 

After working on the code for a while, I noticed that the C and C++ templates were almost exact duplicates in code and that the new test target was somewhat out of place (someone simply copying the template to a folder would not have access to certain targets from within QDMI by default).
Hence, I decided to refactor the templates by combining them into a single template with a small switch for C or C++ and with the respective tests directly embedded in the template.
This makes for a much smaller change and one less place to update on changes.

PR #79, opened 2024-10-29
✨ Docs: Page on Example Device

ystade:
### Issue

To get a quick feeling for QDMI, esp., device developers, they need to dig deep into the code. There are already example implementations for devices in C and C++ but to see those, one needs to inspect the GitHub repository.

### Solution

This PR includes those example implementations of the devices in small snippets into an example page that is part of the documentation.

### Remarks

This PR also tries to finally fix the resizing issue of the Documentation button and fixes some minor issue in the device implementations.


ystade:
Ok, the Documentation button in the README.md is still not really responsive... the property `min-width`is somewhere lost.

burgholzer:
Just fyi: the clang-tidy warnings are popping up here now because this is the first PR after its introduction touching these files.
They should be easily addressed and you can use the last couple of PRs that went into `develop` as inspiration.

burgholzer (approved):
Great stuff. ✨ it's always great to have more documentation that is easily accessible.
I just put on some finishing touches. This is ready to go in once all checks are green ✅

Issue #81, opened 2024-10-30
🎨 Use `size_t` for sizes of buffers and number of returned values


burgholzer:
Quite a lot of the functions within QDMI pass around buffers that are subsequently filled with data. For all of these, a `size` argument is passed in addition to a pointer to the actual buffer.
Right now, the type used for these size arguments is `int`.
However, C/C++ actually has a proper `size_t` type, which is the unsigned integer type returned by the `sizeof` operator.
In order to avoid some of the casts throughout the QDMI implementations, the function signatures should be adapted to use `size_t` instead of `int` for specifying the size of the buffers as well as the number of returned elements from QDMI functions (which is guaranteed to be non-negative).
In addition, the example implementation should be updated to use `size_t` and remove all the unnecessary casts.

PR #82, opened 2024-10-30
✨ Add Readout Format for State Vector and Probabilities

### LINKED ISSUE #80: Add readout function for probabilities to QDMI
Currently there one has the possibility to use the "QDMI_control_readout_raw" function to transfer the readout results via 
QMDI. The function expects integer values for the results, i.e. number of counts per state (int array "num").

For simulated backends it is also possible to calculate the exact probability without having a finite shot number. Therefore it may makes sense to provide the possibility to give probabilities as result instead of counts. This is currently not possible due to the fact that "num" expects integer values. Maybe this can be achieved by providing an additional function (e.g. QDMI_control_readout_probabilities) that expects a double array "num" instead of an integer one, or by adapting the function QDMI_control_readout_raw.

ystade:
### Issue

As discussed in #80, more readout formats are desired, esp., in the case when the QDMI device is a simulator.

### Solution

This PR adds the readout types and implements the corresponding behaviour in both example devices.
```C
QDMI_JOB_RESULT_STATEVECTOR_DENSE,
QDMI_JOB_RESULT_PROBABILITIES_DENSE,
QDMI_JOB_RESULT_STATEVECTOR_SPARSE_KEYS,
QDMI_JOB_RESULT_STATEVECTOR_SPARSE_VALUES,
QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS,
QDMI_JOB_RESULT_PROBABILITIES_SPARSE_VALUES,
```

### DIFF
--- a/examples/device/c/device.c
+++ b/examples/device/c/device.c
@@ -21,7 +21,9 @@ typedef struct QDMI_Job_impl_d {
   QDMI_Job_Status status;
   size_t num_shots;
   char *results;
+  size_t results_length; // includes null terminator
   double *state_vec;
+  size_t state_vec_length;
 } QDMI_Job_impl_t;
 
 typedef struct QDMI_Site_impl_d {
@@ -336,21 +338,23 @@ int QDMI_control_submit_job_dev(QDMI_Job job) {
   size_t num_qubits = 0;
   QDMI_query_device_property_dev(QDMI_DEVICE_PROPERTY_QUBITSNUM, sizeof(size_t),
                                  &num_qubits, NULL);
-  job->results = (char *)malloc(job->num_shots * (num_qubits + 1));
+  job->results_length = job->num_shots * (num_qubits + 1);
+  job->results = (char *)malloc(job->results_length);
   for (size_t i = 0; i < job->num_shots; ++i) {
-    // generate random 5-bit string
-    for (size_t j = 0; j < 5; ++j) {
+    // generate random bitstring
+    for (size_t j = 0; j < num_qubits; ++j) {
       *(job->results + (i * (num_qubits + 1) + j)) = (rand() % 2) ? '1' : '0';
     }
     if (i < job->num_shots - 1) {
       *(job->results + ((i + 1) * (num_qubits + 1) - 1)) = ',';
     }
   }
-  *(job->results + (job->num_shots * (num_qubits + 1) - 1)) = '\0';
+  *(job->results + (job->results_length - 1)) = '\0';
   // Generate random complex numbers and calculate the norm
-  job->state_vec = (double *)malloc((2ULL << num_qubits) * sizeof(double));
+  job->state_vec_length = 2ULL << num_qubits;
+  job->state_vec = (double *)malloc(job->state_vec_length * sizeof(double));
   double norm = 0.0;
-  for (size_t i = 0; i < 1ULL << num_qubits; ++i) {
+  for (size_t i = 0; i < job->state_vec_length / 2; ++i) {
     const double real_part = (((double)rand() / RAND_MAX) * 2.0) - 1.0;
     const double imag_part = (((double)rand() / RAND_MAX) * 2.0) - 1.0;
     norm += real_part * real_part + imag_part * imag_part;
@@ -359,7 +363,7 @@ int QDMI_control_submit_job_dev(QDMI_Job job) {
   }
   // Normalize the vector
   norm = sqrt(norm);
-  for (size_t i = 0; i < 2ULL << num_qubits; ++i) {
+  for (size_t i = 0; i < job->state_vec_length; ++i) {
     // NOLINTNEXTLINE(*-core.UndefinedBinaryOperatorResult)
     job->state_vec[i] = job->state_vec[i] / norm;
   }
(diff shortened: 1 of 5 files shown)


examples/device/c/device.c
```diff
@@ -426,92 +445,254 @@ int QDMI_control_get_data_dev(QDMI_Job job, const QDMI_Job_Result result,
...
+    QDMI_query_device_property_dev(QDMI_DEVICE_PROPERTY_QUBITSNUM, sizeof(int),
+                                   &num_qubits, NULL);
+    // count non-zero elements
+    int count = 0;
+    for (size_t i = 0; i < 1ULL << num_qubits; i++) {
+      if (dense_data[2 * i] != 0.0 || dense_data[(2 * i) + 1] != 0.0) {
+        count++;
+      }
+    }
+    if (data != NULL) {
```
burgholzer: The sparse functions for the state vector and the probability vector still feature the kind of code duplication that was also present in the histogram output before this PR.
Would be great if we could also reduce the duplication here.
See https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/82/commits/3d396d51b3c901602fc3e137ff8cd42fca835145 for more details on what I changed up in the histogram results implementation.

Another important point in that regard: in most of these routines, the call to the `DENSE` variant can be omitted and the `job->state_vec` data can be directly used.

This comment holds for both the C and the C++ device.

burgholzer (changes requested):
Thanks for your work on this 🙏🏼 
I added a bit of testing functionality and fixed a couple of bugs in the implementation.
I also rebased this on top of #84.

As a consequence, clang-tidy does now complain quite a bit more than before 🙃 (mainly cause by the `qdmi::warnings` target now being linked to the devices. However, all of these should be easily addressed.

Finally, the parts that codecov claims are uncovered, are mostly the "failure" branches, where the provided buffers are too small.
It would be great, if we could also add tests for these cases. Especially, to demonstrate the expected behavior.

Given the amount of conversion warnings here, I am thinking whether it would maybe be better to get a fix for #81 in first and rebase on top of that. 
@ystade what do you think? 
I can work on #81 tomorrow.

ystade:
> Thanks for your work on this 🙏🏼 I added a bit of testing functionality and fixed a couple of bugs in the implementation. I also rebased this on top of #84.
> 
> As a consequence, clang-tidy does now complain quite a bit more than before 🙃 (mainly cause by the `qdmi::warnings` target now being linked to the devices. However, all of these should be easily addressed.
> 
> Finally, the parts that codecov claims are uncovered, are mostly the "failure" branches, where the provided buffers are too small. It would be great, if we could also add tests for these cases. Especially, to demonstrate the expected behavior.
> 
> Given the amount of conversion warnings here, I am thinking whether it would maybe be better to get a fix for #81 in first and rebase on top of that. @ystade what do you think? I can work on #81 tomorrow.

Yes, I am in favour of first addressing #81 and rebasing this. This saves a couple of static_casts or warnings.

examples/device/c/device.c
ystade: Is that desired to refer to `job->state_vec` and respective data structures rather than calling the `DENSE` function? Calling the dense function is more generic and also works if one decides to store the results somewhere or somehow different.

burgholzer: I believe that it is fine to use whatever the job stores as it's results. This is up to the device implementer anyway. We have to keep in mind that not all the devices will be able to provide all of the results for a given job.
This will most likely require a way to query, which kind of results are supported by the device and then explicitly requesting them as part of the job (parameters).
But I guess this is something for a different PR.

I'd generally try to keep the example device implementations as simple as possible.

burgholzer: Another argument: I wouldn't like to set the precedent that it is common practice to call the dense methods as part of the sparse methods. The sparse methods are there for a reason: in many practical cases you won't have the memory in all the world to store the dense result, but you might be able to directly report back the sparse result.
Us even being able to implement it in this fashion is simply due to the fact that we are only looking at a small example use case and that we generate a random result in the form of a random dense state-vector.

I would expect that practical implementations of simulators actually would implement this the other way round in many cases. That is, they offer the sparse result and, optionally, a way to create a dense result from that assuming that enough memory is available.

burgholzer:
Alright, I rebased on top of #85 (which addressed #81) and cleaned up the commit history here.
Modulo some code quality improvements, this should be ready to go 🚀

examples/device/c/device.c
ystade: True, agreed.

burgholzer: https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/82/commits/ab65825961ba8e5b7932d96b71692151754550b2 now contains these refactors and considerably shrinks the implementation. I think it looks quite nice now.
Given all tests pass, I'd be up for merging this.

ystade:
I skimmed very quickly over the changes and it looks good to me. If you do not want a rigorous review from me and the quick skimming suffices, it is also a go from my side.

burgholzer:
> I skimmed very quickly over the changes and it looks good to me. If you do not want a rigorous review from me and the quick skimming suffices, it is also a go from my side.

That'a sufficient. Many thanks! 
Everything else can be improved in a follow up 😌

burgholzer (approved):

PR #84, opened 2024-10-31
♻️ Fixing Various Bugs

### LINKED ISSUE #83: ♻️ Tracking Device in Job|Site|Operation and Remove Redundant `device` Parameter
### Issue

When querying information about a `QDMI_Site`or `QDMI_Operation`, everytime the `QDMI_Device` has to be passed to the corresponding function as well even though a site or an operation belongs clearly to one device. The same holds for all function where one passes a `QDMI_Job` together with a `QDMI_Device`.

### Idea

We would define a new type for sites, operations, and jobs that defines the respective concept within a device. The existing types would now be defined in the driver in such a way, that they contain a pointer to the respective concept returned by the device and a pointer to the respective device.

ystade:
This PR was thought to fix #83. As documented there, the initial issue cannot be solved in the proposed way. Hence, this PR just contains some bugfixes that were solved along the way.


test/test_qdmi.cpp
```diff
@@ -221,7 +222,7 @@ measure q -> c;
                                   static_cast<int>(sizeof(int)) * size,
                                   val_vec.data(), nullptr),
             QDMI_SUCCESS);
-  std::map<std::string, int> results;
+  std::unordered_map<std::string, int> results;
```
burgholzer:
```suggestion
  std::map<std::string, int> results;
```
This change will be obsolete based on the changes in #82.

test/test_qdmi.cpp
```diff
@@ -14,6 +14,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include <gtest/gtest.h>
 #include <sstream>
 #include <string>
+#include <unordered_map>
```
burgholzer:
```suggestion
```
This change will be obsolete based on the changes in #82.

burgholzer (dismissed):
Great stuff and nice catches 👍🏼 

I am just going to quickly revert two of the changes because they would most likely cause conflicts with #82.
Otherwise, this can directly go in 🚀 

(I will purposely merge this without squashing to preserve the commit history)

burgholzer (approved):

PR #86, opened 2024-11-03
✨ Use Prefixes for Device Symbols and Types

### LINKED ISSUE #73: 🎨 Prefixing Device Functions
Prefixing the functions of a device would have the following advantages:
- device providers can use their prefix of desire, e.g., `ABC_QDMI_query...` for the company ABC
- easier maintenance and debugging as it is clear in which device an error occurred by the name of the function
- implementations of devices as static libraries where dynamic libraries might not be available

Overall, most changes should involve the driver. However, the device implementations also need to include the header in a way that supports those prefixes.

ystade:
This PR provides a solution to #73. After trying out different options, this has evolved into the most promising one.

It forces device implementations to specify a prefix that is used in front of all functions and types defined by the device. This finally allows multiple devices to be statically linked together into one executable. The static driver implementation shows how this can be accomplished.

Additionally, this PR
- removes two superfluous targets in the CMake configuration
- fixes issues with the template
- automates the template building outside of the project


ystade:
@burgholzer I moved the implementation of the static driver into another PR (#89) so that we can decide independently what we do with it. I think this is ready to go now, I am happy about your review. Note also the additional changes contained here:

> Additionally, this PR
> 
> * removes two superfluous targets in the CMake configuration
> * fixes issues with the template
> * automates the template building outside of the project

include/qdmi/device/types.h
```diff
@@ -0,0 +1,58 @@
...
+// Since this is C code, we cannot use the using keyword.
+// NOLINTBEGIN(modernize-use-using)
+
+/**
+ * @brief Opaque type for a job.
+ * @details A job is a handle to manage the job after it is submitted and to
+ * access the results.
+ *
+ * @par
+ * This type is implemented by the driver.
```
mnfarooqi: Is `QDMI_Job` implemented by a driver or a device?

What is the reasoning behind moving it from `qdmi/common` to `qdmi/device` since it is also used by QDMI client?
If it is because it is implemented by a device, should QDMI_device and QDMI_session also be moved to `qdmi/driver` as they are implemented by a driver?

burgholzer: `QDMI_Job` is implemented by each device. The main reason for moving these device-defined types to their own header is to make the prefixing possible/feasible. Each device will define its own (prefixed) version of these types. On the other hand, the (enum) types in `common/types.h` are not intended to be touched by device implementers (e.g., they should not be adding new enums or enum values).
I agree though, that it could make sense to move the `QDMI_Session` and `QDMI_device` to a `driver/types.h` file.
@ystade what do you think?

ystade: To note that first, @mnfarooqi has definitely a point in that this is wrong in the documentation and should be changed before merging this.

The motivation for the `device/types.h` was to be able to copy the `device.h` and `device/*` in order to have everything that must be prefixed for the respective device implementation.

`QDMI_Session` and `QDMI_device` culd definitely be moved to the `driver`. Then I think the responsibilities should be clear.

burgholzer: > QDMI_Session and QDMI_device could definitely be moved to the driver. Then I think the responsibilities should be clear.

Done in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/86/commits/62ce706294645bc305b6d03febcd0ea7e0766e8a

burgholzer: > To note that first, @mnfarooqi has definitely a point in that this is wrong in the documentation and should be changed before merging this.

Done in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/86/commits/8c3ed5e427a155037f8ebf7dd4fa20189e3b4ccd

burgholzer (approved):
Many thanks @ystade for the great work on this.
This is looking really clean.
I pushed a couple of commits simplifying bits and pieces. Especially, I reduced the code duplication across the CMake files.

This is good to go from my side. If you are fine with it as well, please go ahead and merge :rocket:

PR #89, opened 2024-11-06
✨ Static Device Libraries and Statically Linked Driver

ystade:
This PR contains a statically linked driver. This also allows using of static device libraries. The required preparations have been done by #86.


ystade:
One big issue with this PR is that the existing dynamic driver and the new static driver must be tested independently because they cannot be linked into one executable. Right now—without a separate target for the tests of the static driver—one can simply change the line

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/b16f36d87703c81a95ff274a59fee12f886c2d71/test/CMakeLists.txt#L60

to `dmi::example_static_driver` and the tests will run through as well without any further change.

burgholzer:
> One big issue with this PR is that the existing dynamic driver and the new static driver must be tested independently because they cannot be linked into one executable. 

I think it's even more than just the static driver. 
The whole point of the static driver is to also build the devices themselves as static libraries to get a fully statically-linked binary.
However, currently, the devices are fixed to be dynamic libraries.
I would guess that testing both would either require considerable code duplication or it would effectively require two builds and test runs; one with `-DBUILD_SHARED_LIBS=ON` and one with the setting `OFF`. Depending on whether dynamic or shared libraries are built, the corresponding driver is built and linked.

Let's get #86 in and then rebase this branch on top of develop to get a clean commit history and then see further.

burgholzer:
Closing this for now, but not deleting the respective branch to preserve this proof of concept.

PR #100, opened 2024-11-13
Provision to set session parameters

### LINKED ISSUE #96: User management and access control using sessions
@burgholzer @echavarria-lrz  

`QDMI_session_alloc` is the first function called by a client. However, the function signature doesn't contain anything to pass information about the user/client that is needed for user management and access control when a session is allocated/created. 

Is there any other way for a driver to get this information?

Suggested solution:
The `QDMI_session_alloc` signature can be modified to add a QInfo object as input, which the user can use to pass the required information.

mnfarooqi:
Closes #96 

- Added function `QDMI_session_set_parameter` to set session parameters 
- Added enum `QDMI_SESSION_PARAMETER_T` to be used as parameter to `QDMI_session_set_parameter`
- Added two enum values `QDMI_SESSION_PARAMETER_OWNER` and `QDMI_SESSION_PARAMETER_TOKEN` to `QDMI_SESSION_PARAMETER_T` to be used for access control within a session.

### DIFF
--- a/examples/driver/qdmi_example_driver.cpp
+++ b/examples/driver/qdmi_example_driver.cpp
@@ -271,6 +271,8 @@ int QDMI_session_set_parameter(QDMI_Session session,
   case QDMI_SESSION_PARAMETER_TOKEN:
     session->token = std::string(static_cast<const char *>(value), size);
     return QDMI_SUCCESS;
+  default:
+    break;
   }
   return QDMI_ERROR_NOTSUPPORTED;
 }
--- a/examples/driver/qdmi_example_driver.cpp
+++ b/examples/driver/qdmi_example_driver.cpp
@@ -108,6 +108,8 @@ struct QDMI_Device_impl_d {
  */
 struct QDMI_Session_impl_d {
   std::vector<std::shared_ptr<QDMI_Device_impl_d>> device_list;
+  std::string owner;
+  std::string token;
 };
 
 /// @}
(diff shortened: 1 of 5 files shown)


burgholzer (changes requested):
This is really great! Thanks for the PR 🙏🏼 
I have no comments on the actual additions in the PR. They look very clean.
I have one request though: could we add a default implementation of the new function to the example driver?
The session struct could just add a `owner` and `token` string member and use the new function to set them.
Based on that example implementation, a test could be added for actually using the new method to set these parameters.

I am not quite sure how far we want to take this in the examples, but I could imagine it would be useful if the new session information were actually to be used in the `get_devices` method for selecting which devices to return so that the new values are not only set, but also used in the examples.
Let me know if you think that is taking it too far for now.

mnfarooqi:
> This is really great! Thanks for the PR 🙏🏼 I have no comments on the actual additions in the PR. They look very clean. I have one request though: could we add a default implementation of the new function to the example driver? The session struct could just add a `owner` and `token` string member and use the new function to set them. Based on that example implementation, a test could be added for actually using the new method to set these parameters.
> 
Already implemented will push it in the next commit. 

> I am not quite sure how far we want to take this in the examples, but I could imagine it would be useful if the new session information were actually to be used in the `get_devices` method for selecting which devices to return so that the new values are not only set, but also used in the examples. Let me know if you think that is taking it too far for now.
>
Working on it too.

burgholzer:
> > This is really great! Thanks for the PR 🙏🏼 I have no comments on the actual additions in the PR. They look very clean. I have one request though: could we add a default implementation of the new function to the example driver? The session struct could just add a `owner` and `token` string member and use the new function to set them. Based on that example implementation, a test could be added for actually using the new method to set these parameters.
> 
> Already implemented will push it in the next commit.
> 
> > I am not quite sure how far we want to take this in the examples, but I could imagine it would be useful if the new session information were actually to be used in the `get_devices` method for selecting which devices to return so that the new values are not only set, but also used in the examples. Let me know if you think that is taking it too far for now.
> 
> Working on it too.

Awesome! Many thanks 🙏🏼

burgholzer (approved):
Great stuff! Thanks for taking care of this so quickly.
Feel free to merge the PR if you are happy with it 🙌

PR #101, opened 2024-11-18
🧪 Test Installation of QDMI

### LINKED ISSUE #98: 🧪 Test Installation of QDMI
So far there is no tests that tests whether all tests also pass when QDMI is installed instead of build in a build directory.

ystade:
This PR tries to add the missing test for the installation of QDMI pointed out in #98.

### DIFF
--- a/.github/workflows/reusable-cpp-install-tests.yml
+++ b/.github/workflows/reusable-cpp-install-tests.yml
@@ -36,7 +36,7 @@ jobs:
         uses: Chocobo1/setup-ccache-action@v1
         with:
           prepend_symlinks_to_path: false
-          override_cache_key: c++-tests-install-${{ inputs.os }}
+          override_cache_key: c++-tests-install-${{ matrix.os }}
       # set up mold as linker for faster C++ builds
       - name: Set up mold as linker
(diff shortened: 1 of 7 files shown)


.github/workflows/reusable-cpp-install-tests.yml
```diff
@@ -0,0 +1,63 @@
...
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+      # set up ccache for faster C++ builds
+      - name: Setup ccache
+        uses: Chocobo1/setup-ccache-action@v1
+        with:
+          prepend_symlinks_to_path: false
+          override_cache_key: c++-tests-install-${{ inputs.os }}
```
burgholzer:
```suggestion
          override_cache_key: c++-tests-install-${{ matrix.os }}
```

burgholzer (approved):
This is looking great! Many thanks for adding these tests and fixing the general library structure for installation.
I just spotted a simple to fix problem in the CI workflow cache key that should be fixed by adopting the inline suggestion.
Let's get this in 🚀

PR #120, opened 2024-12-04
🐛 Fix an inconsistency in the docs about coupling maps

### LINKED ISSUE #119: Type of coupling map
If I'm not mistaken, the [type of the coupling map of the example device](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/develop/examples/device/cxx/device.cpp#L145) does not comply with the [type specified in enums.h](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/develop/include/qdmi/common/enums.h#L45). We are currently working on implementing the query interface and were wondering whether the coupling map should be a list of `int`s or a list of `QDMI_Device_Site`s.

ystade:
This PR fixes an inconsistency that was present in the documentation of the coupling map parameter, and that was reported in #119.

Fixes #119

### DIFF
--- a/include/qdmi/common/enums.h
+++ b/include/qdmi/common/enums.h
@@ -42,13 +42,15 @@ enum QDMI_DEVICE_PROPERTY_T {
   /// `size_t` The number of qubits in the device.
   QDMI_DEVICE_PROPERTY_QUBITSNUM,
   /**
-   * @brief `int*` (int list) The coupling map of the device.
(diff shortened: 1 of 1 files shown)


burgholzer (approved):
LGTM 👍🏼 Thanks for the quick fix!

echavarria-mqv:
The original issue was created just 30 minutes ago, if this isn't top-tier customer service, I’m not sure what is ;)

PR #122, opened 2024-12-04
🎨 Stability improvements for custom enum parameters

### LINKED ISSUE #118: 🐛 Compatibility Concerning Custom Parameters
It just came up in a discussion with @burgholzer that most (all) of the enumerations used to define the different parameters, e.g., `QDMI_DEVICE_PROPERTY_T`, define a couple of custom values. Those enable device implementers to quickly test the exposure of properties that are not yet included in the QDMI without requesting a change of QDMI.

The issue with those custom parameters is that they are supposed to follow after all regular enumeration values. The last member of the enumeration is then a value that defines the number of elements in the enumeration and is used for bounds checking, i.e., whether the passed enumeration value is valid.

When an upgrade of QDMI defines new enumeration values in some of those enumerations, those custom parameters are pushed back, and their value increases. Hence, old QDMI device implementations that support a custom parameter may not handle this parameter correctly anymore. Since the concrete integer value behind the enumeration constant is assigned to another identifier now. In the compiled library, however, not the identifier of the enumeration members is used but instead the actual value of those constants (which now became another meaning).

The proposal to solve this issue would be, to assign the maximal possible integer values to the custom parameter enumeration values in order to
1. fix the value of those custom parameters across versions of QDMI, and
2. enable better validation of the passed parameter values.
For the second point, note that the `...MAX` member of each enumeration should denote the number of __regular__ members, excluding the custom members of an enumeration. Hence, the custom parameters will always lay outside of the allowed range for parameter values. This is sensible, as no device is required to implement those custom parameters. They are only meant for custom experimental usage.

ystade:
Fixes #118
by moving the custom parameters to the end of structs and assigning a fixed high value to them that must not be changed.

This PR also refactors all `..._CUSTOM_[int]` to `..._CUSTOM[int]` without the last underscore to follow the naming conventions in QDMI.

Finally, it improves upon the docstrings of certain methods.


burgholzer (changes requested):
I think we can safely ignore the clang-tidy warnings file-wide as this is a design decision on purpose.

Tests still fail at the moment unfortunately.

ystade:
> I think we can safely ignore the clang-tidy warnings file-wide as this is a design decision on purpose.
> 
> Tests still fail at the moment unfortunately.

@burgholzer, I actually think the clang tdiy warning has a point and I find it clearer when all constants have assigned integer values. I think that makes our intend even clearer.

Hopefully, the tests pass now.

burgholzer (approved):
LGTM now. Let's get this in 🚀

PR #126, opened 2024-12-09
✨ Provide an option to specify the supported QIR profile

### LINKED ISSUE #115: ✨ Provide an option to specify the supported QIR profile
Right now, there is no way for a device to specify which QIR profile it supports.
It should be easy enough to add this as another device property that has a string value.
As far as I am aware, the only possible values for that string at the moment would be `"base_profile"` and `"adaptive_profile"`.

This is slightly in contrast to how we handle the support for OpenQASM where we offer separate enum values for OpenQASM 2 and OpenQASM 3 support.
However, I do think this makes sense since OpenQASM 3 isn't really a superset of OpenQASM 2 in the way that the adaptive QIR profile is a superset of the base profile.

ystade:
This PR provides the solution as discussed in issue #115. Hence, it is possible to first ask a device whether a certain job format is supported before actually submitting the program in that format.

Fixes #115

### DIFF
--- a/examples/device/c/device.c
+++ b/examples/device/c/device.c
@@ -303,18 +303,23 @@ int C_QDMI_query_operation_property_dev(C_QDMI_Operation operation,
 int C_QDMI_control_create_job_dev(const QDMI_Program_Format format,
                                   const size_t size, const void *prog,
                                   C_QDMI_Job *job) {
+  if (format != QDMI_PROGRAM_FORMAT_QASM2 &&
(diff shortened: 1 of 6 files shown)


burgholzer (approved):
Many thanks for creating this pull request.
I quickly went over everything and slightly adjusted the API documentation.
From my perspective, everything is in order, but could you please review the updated API descriptions and verify if my modifications make sense?
If so, please go ahead and merge. 🚀

burgholzer (approved):
LGTM ✅  many thanks for the additional fixes

PR #129, opened 2024-12-13
✨ Add `ID` as a Site Property

### LINKED ISSUE #128: QDMI_Site unique identifier missing
A qubit is represented as `QDMI_Site` by an opaque pointer to be implemented by a `device`. The only two properties defined for `QDMI_Site` are `T1` and `T2` times.
Shouldn't there be a property (e.g. `QDMI_SITE_PROPERTY_ID` or `QDMI_SITE_PROPERTY_INDEX`) that uniquely identifies a `QDMI_Site` that can be used by a mapper to map circuit qubits to device qubits?

PS: I went over the example implementation and `QDMI_Site` has a single member variable `id` and I think the example `tool` works because of this. However, we cannot assume the same for an opaque pointer implemented by a device. 

Did I get it right or did I miss something?

@burgholzer @ystade

ystade:
As pointed out in #128, the software stack wants to address specific sites on the device in the compiled QASM/QIR/... programs. Hence, there must be a way for the client to know the id of a site. For that, this PR add the new `QDMI_SITE_PROPERTY_ID`.

Fixes #128

### DIFF
--- a/include/qdmi/common/enums.h
+++ b/include/qdmi/common/enums.h
@@ -98,8 +98,20 @@ enum QDMI_DEVICE_STATUS_T {
 
 /// Enum of the site properties that can be queried.
 enum QDMI_SITE_PROPERTY_T {
-  /// `size_t` The id to identify the site in the circuit.
-  /// @details For example, `q[1]` in a QASM code identifies the site with id 1.
+  /**
+   * @brief `size_t` The id to identify the site in the circuit.
+   * @details For example, `$0` in a QASM code identifies the (physical) site
+   * with id 1. Note, those physical qubits in the QASM language should not be
+   * confused with the virtual qubits used more frequently. Those are allocated
+   * via `qreg q[3]`, for example, and then used in the circuit as, e.g.,
+   * `h q[0]`.
+   *
+   * @par
+   * On the QIR side, there is also the distinction between dynamically
+   * allocated and static hardware qubits. Here, only the latter refer to the
+   * physical ids of sites. E.g., `ptr inttoptr (i64 1 to ptr)` refers to the
+   * physical site with id 1.
+   */
   QDMI_SITE_PROPERTY_ID = 0,
   QDMI_SITE_PROPERTY_TIME_T1 = 1, ///< `double` The T1 time of a site in µs.
   QDMI_SITE_PROPERTY_TIME_T2 = 2, ///< `double` The T2 time of a site in µs.
--- a/include/qdmi/common/enums.h
+++ b/include/qdmi/common/enums.h
@@ -98,8 +98,11 @@ enum QDMI_DEVICE_STATUS_T {
 
 /// Enum of the site properties that can be queried.
 enum QDMI_SITE_PROPERTY_T {
-  QDMI_SITE_PROPERTY_TIME_T1 = 0, ///< `double` The T1 time of a site in µs.
-  QDMI_SITE_PROPERTY_TIME_T2 = 1, ///< `double` The T2 time of a site in µs.
+  /// `size_t` The id to identify the site in the circuit.
+  /// @details For example, `q[1]` in a QASM code identifies the site with id 1.
+  QDMI_SITE_PROPERTY_ID = 0,
+  QDMI_SITE_PROPERTY_TIME_T1 = 1, ///< `double` The T1 time of a site in µs.
+  QDMI_SITE_PROPERTY_TIME_T2 = 2, ///< `double` The T2 time of a site in µs.
   /**
    * @brief The maximum value of the enum.
    * @details It can be used by devices for bounds checking and validation of
@@ -107,7 +110,7 @@ enum QDMI_SITE_PROPERTY_T {
    * enum besides the custom members and must be updated when new members are
    * added.
    */
-  QDMI_SITE_PROPERTY_MAX = 2,
+  QDMI_SITE_PROPERTY_MAX = 3,
   /**
    * @brief This property is reserved for a custom property.
    * @details The meaning and the type of this property are defined by the
--- a/include/qdmi/common/enums.h
+++ b/include/qdmi/common/enums.h
@@ -101,10 +101,19 @@ enum QDMI_SITE_PROPERTY_T {
   /**
    * @brief `size_t` The id to identify the site in the circuit.
    * @details For example, `$0` in a QASM code identifies the (physical) site
-   * with id 1. Note, those physical qubits in the QASM language should not be
+   * with id 0. Note, those physical qubits in the QASM language should not be
    * confused with the virtual qubits used more frequently. Those are allocated
    * via `qreg q[3]`, for example, and then used in the circuit as, e.g.,
    * `h q[0]`.
+   *
+   * @par
+   * Additionally, the qubits in a QASM program that contain exactly one
+   * register with exactly the same number of qubits as the device has sites is
+   * also handled as physical qubits. For example for a device with three qubits
+   * and a QASM program `qreg q[3]; h q[0]; cx q[0], q[1];`, the operations are
+   * indeed performed on the physical sites with ids 0 and 1. Note here, that
+   * this is not the intended use of virtual qubits in QASM, but because of
+   * their wide adoption in the community, this behavior is also supported.
    *
    * @par
    * On the QIR side, there is also the distinction between dynamically
--- a/test/test_qdmi.cpp
+++ b/test/test_qdmi.cpp
@@ -137,12 +137,13 @@ TEST_P(QDMIImplementationTest, QueryGatePropertiesForEachGate) {
 }
 
 TEST_P(QDMIImplementationTest, QuerySiteProperties) {
-  // for every gate in the gate set, query the duration of the gate
+  // get all sites and check whether their count is equal to the number of
+  // qubits
   const auto fomac = FoMaC(device);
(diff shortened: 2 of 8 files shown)


include/qdmi/common/enums.h
```diff
@@ -98,16 +98,19 @@ enum QDMI_DEVICE_STATUS_T {
 
 /// Enum of the site properties that can be queried.
 enum QDMI_SITE_PROPERTY_T {
-  QDMI_SITE_PROPERTY_TIME_T1 = 0, ///< `double` The T1 time of a site in µs.
-  QDMI_SITE_PROPERTY_TIME_T2 = 1, ///< `double` The T2 time of a site in µs.
+  /// `size_t` The id to identify the site in the circuit.
+  /// @details For example, `q[1]` in a QASM code identifies the site with id 1.
```
burgholzer: I'd prefer to use the `/**` comment syntax for multi-line comments to stay consistent with the rest.

The details comment isn't quite accurate. If a quantum circuit defines a register `qreg p[1]` before the register declaration of `q`, all indices in q need to be shifted by 1.
I would rather like to draw the analogy with OpenQASM's physical qubits concept (https://openqasm.com/language/types.html#physical-qubits) or QIR's static addressing mode (casting ints to pointers).
An interesting bit in the OpenQASM spec is that IDs need not be consecutive.
I believe this might invalidate the test that you added here.

In my opinion, QDMI is not at all point where it could enforce QASM and QIR programs to be single-register. At least the current interface has no say in this.

test/test_qdmi.cpp
```diff
@@ -136,6 +136,19 @@ TEST_P(QDMIImplementationTest, QueryGatePropertiesForEachGate) {
   }
 }
 
+TEST_P(QDMIImplementationTest, QuerySiteProperties) {
+  // for every gate in the gate set, query the duration of the gate
```
burgholzer: This is not what the test does 😉

burgholzer (changes requested):
Thanks for adding this so quickly!
I have a couple of concerns that I believe need to be addressed before adding this. 

One thing I didn't find a good place to mention inline is that we most likely want this to be a required property of a site. Every device will need to provide this property for its sites, otherwise the use case from the issue would not work everywhere.

Furthermore, there needs to be quite some documentation about the expected mapping of program qubits to sites.
For OpenQASM, this is moderately easy as the standard only allows to allocate but not to deallocate qubit registers. That means one can simply index the qubits based on the sequence on which the registers are declared.
For QIR, this is definitely harder when not enforcing the QIR program to be at the hardware level with all qubits known and no register allocations. Because QIR does allow to deallocate qubit registers. So the indexing becomes way harder.

Just a thought to spark some discussion 💭
Would it make sense to query sites from devices via an index? That is, one could query a device with 3 qubits "give me the site with index 0, with index 1, and with index 2"

include/qdmi/common/enums.h
ystade: Good point, I extended the documentation here and mentioned hardware qubits both in QIR and QASM.

ystade:
Reply on the first part: When referring to hardware qubits, i.e., the id of sites, only the hardware profile of the respective language makes sense. Dynamically allocated can still be placed freely on the device. Otherwise, the concept of dynamically allocating qubits does not make sense.

ystade:
Reply on the second part: I was even thinking, whether the sites should only be identified by their id. This means, there is no function `...get_sites` anymore and when querying a site, one would pass the id to the site to the `...query_site_property` function. We had that at some point and found the solution with the opaque pointers more elegant and I am not really in favour of switching back but wanted to bring that up here.

Furthermore, one could actually avoid that sites must have an id. Instead, one could introduce a `...DEVICE_PARAMETER`, that allows to set a mapping from `int`s (IDs) to `QDMI_Site`s. This would for example look like this: `QDMI_device_set_parameter(device, QDMI_DEVICE_PARAMETER_SITE_MAPPING, sizeof(...), {2, sites[3]})`. This line of code would map the index 2 that appears in a quantum program to the fourth site (`sites[3]`) returned by `...get_sites`. I am neither in favour of this option but also wanted to show that there are alternatives.

burgholzer:
> Reply on the first part: When referring to hardware qubits, i.e., the id of sites, only the hardware profile of the respective language makes sense. Dynamically allocated can still be placed freely on the device. Otherwise, the concept of dynamically allocating qubits does not make sense.

That implies that any program that is not written in terms of hardware qubits (even if it is a single-register QASM program) is not guaranteed to be executed on the qubits indexed in the program.
While I agree that this should be the eventual goal, in the early adoption phase of QDMI, very generic programs will be sent across the interface.
Should we say something explicit about this? For example that programs that use hardware qubits and programs that use single registers (named "q" and having the size of the device) are expected to run directly on the respective qubits. And maybe even that a device might reject other programs. Not too sure about that though. In the end, the interface should remain rather generic.

ystade:
We could add this feature as a note in the documentation to make clear that this is not really the intended use but we support it as of now.

burgholzer:
> Reply on the second part: I was even thinking, whether the sites should only be identified by their id. This means, there is no function `...get_sites` anymore and when querying a site, one would pass the id to the site to the `...query_site_property` function. We had that at some point and found the solution with the opaque pointers more elegant and I am not really in favour of switching back but wanted to bring that up here.

The pro of this would be that it is truly enforced that sites need an ID.
I like that aspect.

It would force the device to use consecutive IDs though (which is slightly more restrictive than the OpenQASM spec for example). Rigetti's device indexing is a good example of this. They have a ring-based architecture, where 8-qubit rings are stitched together at their boundaries. Their indexing scheme then is `YYX` where `YY` is the ring index and `X` is the index within the ring (0-7).
I am on the fence regarding whether we want to support that or just require consecutive indexing.

Indexing by integers is slightly more convenient than passing around QDMI_Site handles.

> Furthermore, one could actually avoid that sites must have an id. Instead, one could introduce a `...DEVICE_PARAMETER`, that allows to set a mapping from `int`s (IDs) to `QDMI_Site`s. This would for example look like this: `QDMI_device_set_parameter(device, QDMI_DEVICE_PARAMETER_SITE_MAPPING, sizeof(...), {2, sites[3]})`. This line of code would map the index 2 that appears in a quantum program to the fourth site (`sites[3]`) returned by `...get_sites`. I am neither in favour of this option but also wanted to show that there are alternatives.

This is comparable to what we do in MQT Core via the `initialLayout` attribute of the quantum circuit class and similar to what qiskit does with their layout property on circuits.
I believe most mature quantum circuit IRs have some notion of this kind of dictionary. However, the circuit formats don't preserve that kind of information.
What I am trying to say here: this kind of information is not standardized (yet).

I think this would work, but would be cumbersome to use and to get right (in documentation and execution).

ystade:
So, @burgholzer, what is your favoured way to go here? Also others, e.g., @mnfarooqi, are welcome to jump in here.

My preference would be way that this PR follows (after the clang-tidy warnings are fixed)

burgholzer:
> So, @burgholzer, what is your favoured way to go here? Also others, e.g., @mnfarooqi, are welcome to jump in here.
> 
> 
> 
> My preference would be way that this PR follows (after the clang-tidy warnings are fixed)

I think I like the way that this PR is handling the situation. I have some nuanced comments about the documentation, which I'd like to slightly tweak. However, I am taking off in a couple minutes (🛫) and just have my phone at hand. I'll try to provide more details in the evening.

Do we want to additionally or exclusively allow to query site properties by ID? 
I could see some benefits in that.

Do we really want to enforce that the IDs are consecutive? I am fine with that. Especially since it makes some handling easier.
However, this requirement should be made even clearer in the documentation.

ystade:
I see the querying sites by id as an exclusive swap out for the current solution, I would not like to maintain to in parallel. I think this is just a source for errors.

ystade:
Regarding the consecutiveness of IDs. This is *not* enforced in the tests for the template only in the tests for the implementation of the example devices within QDMI. However, even there we can remove it. I would not force a device to have consecutive IDs. As soon as, one qubit/site fails and is out of service you will not have consecutive site IDs anymore.

ystade:
Somehow connected, there is also the question how a client knows how to interpret the measurement results returned from a device, i.e., which data belongs to which qubit in the results.

burgholzer:
> Somehow connected, there is also the question how a client knows how to interpret the measurement results returned from a device, i.e., which data belongs to which qubit in the results.

That's an even different story. Measurement results do not belong to qubits. Imagine a dynamic quantum circuit that repeatedly measures a single-qubit over the course of the computation.
The results are stored in classical registers (in OpenQASM that is very literal; in QIR results are also stored in classical bits).
I would apply a very similar line of reasoning here: bits are returned in order of their declaration. So for classical registers `c` and `d`, each with two bits, the results vector would have four elements with c[0] at 0, c[1] at 1, d[0] at 2 and d[1] at 3.

I am not 100% certain how that translates to QIR.

mnfarooqi:
I think the current PR addresses the problem at hand. Regarding ID sequencing, it is good to have it, but not a must, as tool developers can still deal with it.

One thing for documentation is to clarify what ID means. Is this a global ID across all registers and "cores" or is it an ID within a register or within a "core". 

The solution can evolve over time as hardware architectures and IRs evolve.

burgholzer:
Closing this in favor of #132 which includes all the relevant changes from here and more extensive documentation.

PR #132, opened 2025-01-02
✨ The first official version

### LINKED ISSUE #106: 📝 add documentation page for client implementation
I think we should consider adding the example of a `QDMI client`.

### LINKED ISSUE #108: 🔬 Support for calibration routines
The `v0.1` version of QDMI had dedicated routines for calibration
```c++
int QDMI_device_quality_check(QDMI_Device dev, double *result);
int QDMI_device_quality_limit(QDMI_Device dev, double *result);
int QDMI_device_quality_calibrate(QDMI_Device dev);
```

These have been purposely removed as part of the transition towards `v1` due to their limited usability.
However, calibration is an important aspect of integrating QC systems into HPC environments.
Hence, QDMI should allow to trigger calibration for devices.

@martin-knudsen: Just putting in motion the (re-)integration of this feature as part of the delayed deliverable on "continuous re-calibration" for MUNIQC-SC.

### LINKED ISSUE #109: 🪙 Dealing with API tokens
There is currently no defined way to deal with API tokens that may be needed to interact with the underlying API of a device.

Our (planqc's) device needs an API token for job submission. For now, we have hard-coded the API token in our shared library, but this means that our shared library breaks when the token expires.

If a device needs a token only for job submission, it can, in principle, be set with `QDMI_control_set_parameter`. However, devices may also want to use a token for the query interface.

### LINKED ISSUE #117: 🐛💥 Device implementations only have global state
## Issue description

This issue popped up while discussing the implementation of the changes proposed in #109 with @ystade.
In that issue, the possibility of adding a `QDMI_control_set_device_parameter` method has been brought up, e.g., to store an access token or a base URL.
The problem now arises if two users who have access to the same device handle call the new setter with their respective tokens.
Since the device implementation has no way of knowing from where it is called (it has no back reference to a session or a `QDMI_Device_impl_t` struct from the driver. Hence, there is no way for it to distinguish the two calls. After all, the device simply provides a definition of the setter function that would update some _global state_.
Hence, the second setter call would effectively overwrite the previous call.
Thus, at the moment, QDMI cannot support this multi-user environment adequately.
Given the importance of this use case, a change of the interface specification is necessary.

## Proposed Solution

To allow multiple-user scenarios, QDMI Device implementations need to be able to maintain some sort of _state_. A client would still only receive a `QDMI_Device` handle. However, that handle would internally (i.e., in the driver) maintain a session-specific `QDMI_Device_State` handle in addition to the connection to the shared library symbols (i.e., the device implementation, which is independent of the session). 
The `QDMI_Device_State` struct would be defined by the device itself. The device interface would gain the following additional enumeration and methods
```c++
enum QDMI_DEVICE_STATE_PARAMETER_T {
  QDMI_DEVICE_STATE_PARAMETER_TOKEN, ///< `char*` (string) Token used for API calls in the device implementation.
  QDMI_DEVICE_STATE_PARAMETER_BASEURL, ///< `char*` (string) baseURL used as an API endpoint
  /**
   * @brief This value is reserved for a custom parameter.
   * @details The meaning and the type of this parameter are defined by the
   * device.
   */
  QDMI_DEVICE_STATE_PARAMETER_CUSTOM1,
  ...
}

typedef enum QDMI_DEVICE_STATE_PARAMETER_T QDMI_Device_State_Parameter;

/**
 * @brief Allocate a new QDMI device state.
 * @details The returned handle can be used in subsequent calls to @ref
 * QDMI_device_state_set_parameter
 * @param[out] state a handle to the device state that is allocated. The state
 * must be freed by calling @ref QDMI_device_state_free when it is no longer needed.
 * @return @ref QDMI_SUCCESS if the state was allocated successfully.
 * @return @ref QDMI_ERROR_OUTOFMEM if memory space ran out.
 * @return @ref QDMI_ERROR_FATAL if an unexpected error occurred.
 */
int QDMI_device_state_alloc(QDMI_Device_State* state);

/**
 * @brief Set a parameter for a device state.
 * @details The parameter can be one of the predefined parameters in 
 * @ref QDMI_Device_State_Parameter.
 * @param[in] state refers to the state to set the parameter for.
 * @param[in] param the parameter to set.
 * @param[in] size the size of the value in bytes.
 * @param[in] value the value to set the parameter to.
 * @return @ref QDMI_SUCCESS if the parameter was set successfully.
 * @return @ref QDMI_ERROR_INVALIDARGUMENT if the device state does not exist, is in
 * an invalid state or the parameter is invalid.
 * @return @ref QDMI_ERROR_NOTSUPPORTED if the device does not support the
 * parameter.
 * @return @ref QDMI_ERROR_FATAL if the parameter could not be set.
 */
int QDMI_device_state_set_parameter(QDMI_Device_State state, QDMI_Device_State_Parameter parameter, size_t size, const void *value);

/**
 * @brief Free a QDMI device state.
 * @details This function frees the memory that was allocated for the state.
 * Accessing a (dangling) handle to a state after it was freed is undefined behavior.
 * @param[in] state the state to free.
 */
void QDMI_device_state_free(QDMI_Device_State state);
```

💥 In addition to that, and this is where it gets slightly annoying, most (if not all) functions that are part of the device interface will require an additional `QDMI_Device_State state` parameter (supposedly as the first parameter in the respective methods). Devices can then use this state (e.g., an API token and/or endpoint) in the individual methods (e.g., for querying and/or submitting jobs).
The precise list of methods that need the additional parameter is still to be decided in the actual implementation.

@AdVetter @denialhaag for awareness. Comments/feedback welcome!

### LINKED ISSUE #123: 🍱 Improve clarity of QDMI flow diagram
<p align="center">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/Munich-Quantum-Software-Stack/QDMI/develop/docs/_static/qdmi_dark.svg" width="80%">
    
  </picture>
</p>

I have always had some reservations about the diagram shown in the README file (and in flyers, posters, presentations, etc.). The loops going back to the beginning of both optimization stages make it seem like the design space exploration happens during the just-in-time compilation, but there are a few issues with that approach. I would suggest to remove the loops and present an acyclic diagram instead.

### LINKED ISSUE #125: ♻️ Streamline QDMI Driver Authentication by Requiring Token in Session Initialization
A QDMI driver should be exchangeable, i.e., any client compliant with the QDMI specification should work with any driver implementation.
At the moment, it's up to the driver to decide whether to require a "token" and/or "owner" in order to get devices from a session.
There is no way for the client to tell what is actually needed prior to calling the `QDMI_session_get_devices` function and when that function returns `QDMI_STATUS_PERMISSIONDENIED` there is no way for the client to tell why.
In order to streamline the authentication process between the client and the driver, the proposal would be to require a token in the `QDMI_session_initialize` function and remove the `QDMI_SESSION_PARAMETER_TOKEN` and `QDMI_SESSION_PARAMETER_OWNER` enum values.
As a result, it is clear to the client that a token is required for creating a session.
Additionally, it is clear for the driver that it should expect to be handed a token.

More precisely, the `QDMI_session_initialize` function signature would be changed to
```c++
int QDMI_session_initialize(const char* token, size_t size, QDMI_Session* session)
```

@mnfarooqi @echavarria-lrz do you see any problem with the above proposal? In particular, do you think it is sufficient to rely on a token for client authentication in all cases?

### LINKED ISSUE #128: QDMI_Site unique identifier missing

### LINKED ISSUE #131: Implementation guidelines for functions that take a `size_ret` argument
It is necessary for functions that retrieve data to allocate the correct size. The example device implements this by calling the function twice, once to get the size using `size_ret` and then calling again passing a pointer with the correct amount of memory allocated.

To make it clear to the device developers to implement the expected functionality for functions with `size_ret` argument, it may be explicitly mentioned that passing `data`/`value` as `NULL` and `size` as `0` and `size_ret` as not `NULL` is not a `QDMI_ERROR_INVALIDARGUMENT`, but the function should return the actual size of the data in `size_ret` and ignore `size` and `data`/`value`.

For example, `QDMI_control_get_data_dev` is missing such an explanation.

burgholzer:
✨ Starting the new year with a bang 💥
 
This PR contains the work of the last couple of weeks by @ystade and myself on refining QDMI based on the discussions we had at the end of last year. 
Given the discussions in #109, #117 and #125, it was clear that breaking changes were inevitable.
And this PR definitely breaks things. **However**, it also fixes and improves literally all aspects of QDMI. In particular:

- The interfaces have been consolidated and restructured. There are now two clear interfaces: the "device interface" (contained in the `qdmi/device.h` header) and the "client interface" (contained in the `qdmi/client.h` header). These two interfaces connect three entities: devices, clients, and a driver. This distinction and separation of concerns has been made more explicit in the interface itself as well as in its documentation.
- In addition to the regular `QDMI_Session` concept for establishing a connection between a client and a driver, a new `QDMI_Device_Session` concept was introduced that accomplishes a similar connection between the driver and a device. This resolves #109 and #117, based on the discussion in these issues, as it allows the driver to establish separate connections to a device for each client. The respective sessions may be used to authenticate with the device or to potentially switch the API endpoint used in the device implementation without having to recompile the device implementation.
- Resolves #125 by streamlining authentication and session handling. Based on the discussions over there, the new expected workflow with sessions is to first create them, then set certain parameters, and, finally, to initialize them. A session may only be used once it is successfully initialized. Any kind of authentication is handled in the new initialization routines.
- Resolves #128 and supersedes #129 by adding a unique identifier to QDMI sites. Significant effort has been put into clarifying the respective semantics around site IDs and qubits used in programs submitted to the devices.
- Many functions have been renamed to follow a consistent naming scheme throughout the whole interface. Functions are now generally prefixed by the type they operate on (for example, `QDMI_device`; typically the type of the first parameter), followed by the action to be performed (for example, `query`), and suffixed by the particular thing acted on (for example, `site_property`). This style is now consistently used throughout the whole interface.
- The `QDMI_device_get_sites` and `QDMI_device_get_operations` routines have been refactored to become device properties that can be queried just as every other property. This helps to keep the interface more compact and the semantics more uniform.
- Based on discussion around pulse-level control, the property query function for operations has received additional parameters for querying properties of parametrized operations depending on their parameter values. This allows devices to expose almost arbitrary operations and their properties. Effectively, this should allow realizing pulse-level QDMI devices.
- Resolves #108 by adding the respective `calibration` program format and `needs calibration` device property as discussed in that issue.
- Resolves #123 by adjusting the figure as discussed in that issue.
- Resolves #131 by adding extensive API documentation including examples.
- Resolves #106 by significantly expanding on the client interface documentation.
- The API documentation has been meticulously cleaned up and extended wherever possible.
- All example devices, the example driver, as well as the example fomac and the example tool have been updated based on the above changes.
- All tests have been updated based on the above changes.
- ... (I probably forgot a couple of things)

Overall, I believe the list above illustrates the significant progress made since last year's discussions. In particular, this PR closes all open issues requiring breaking changes that we are aware of. Based on that and the level of attention paid to every little detail in this PR, I am confident that, once merged, this will be the first official stable version of QDMI.

> [!NOTE]
> This is a huge pull request. 
> However, given the sheer scope of the changes in this PR, it is not easily possible to split it into smaller chunks. Splitting it up in retrospect would cause significant churn and lead to further delays in the official first release. 
> During development, we took extra care to make small-grained commits so that the commits themself allow to follow the developments and the individual changes.
> As such, the idea would be to _not_ squash the commit history of this PR before merging, but to perform a regular merge that preserves the individual commits.
> Any potential reviewer is advised to look through the git diff in the `examples` folder as this will nicely illustrate the changes on the device and on the client side.

One of the only questions left is the version number we want to use. The two options I see are `v1.0.0` or `v2.0.0`.
Given how we have not made a public 1.0 release yet, it would be justified to call the release `v1`. However, given how people have started to implement QDMI as it is currently on the `develop` branch and how significant the changes in this PR are to the interface, one might rather argue that we should tag the current version on the `develop` branch as `v1` and the version after merging this PR as `v2`. In my personal opinion, that would be the cleaner solution, but I am open to the opinion of others here.
In any case, the version currently on the develop branch will no longer be supported once this PR is merged.


burgholzer:
Alright, it took a while, but the entire CI is as green as it gets ✅ 
I also added a nice little feature to build documentation previews for PRs so that people can have a look at the documentation before it is merged, without having to build it locally.

The preview for this PR is available at https://munich-quantum-software-stack.github.io/QDMI/pr-preview/pr-132/
This is probably the best way of previewing the changes from this PR.

I will keep this PR open for comments until sometime next week. If nothing big pops up until then, I will go ahead and merge the PR and proceed with the release.

I am explitly tagging a couple of people here that were involved in the related discussions.
@ystade @echavarria-lrz @mnfarooqi @AdVetter @denialhaag

robertwille:
Wow, this is truly impressive!👍 Huge thanks @burgholzer and @ystade for all your efforts on this essentiell component!🙏

ystade:
Regarding the versioning: I would actually be in favor of v1.1.0. I agree that within the interface we made a lot of changes. However, from the perspectives of users of QDMI, e.g., device implementers, I think the changes are still moderate. In the example device we only moved some pieces of code and renamed a couple of functions but not more. Additionally, I am hesitant to leave version v1.0.0 behind because it is not working (see all related issues). In my opinion, we should only start with a new version when we have fixed the old one such that it is working, which is not true for v1 right now.

ystade (approved):
This PR comes with plenty of changes but I think all of them for the good. Everything looks good to me.

burgholzer:
> Regarding the versioning: I would actually be in favor of v1.1.0. I agree that within the interface we made a lot of changes. However, from the perspectives of users of QDMI, e.g., device implementers, I think the changes are still moderate. In the example device we only moved some pieces of code and renamed a couple of functions but not more. Additionally, I am hesitant to leave version v1.0.0 behind because it is not working (see all related issues). In my opinion, we should only start with a new version when we have fixed the old one such that it is working, which is not true for v1 right now.

Totally agree with the reasoning. Adjusted the version number in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/132/commits/d64b4eeab54c4e786aee0ce126eaacadfa247656.
I will tag https://github.com/Munich-Quantum-Software-Stack/QDMI/commit/fa142a656bbed226ba4e38051d00931f9656c2e9 as `v1.0.0` since it was originally intended as such and no _significant_ (breaking) changes have been committed to the `develop` branch since then. Most likely, I will also create an official GitHub release for that so that it does not look as odd when the releases start with `v1.1.0`.
Then, I'll merge this PR and tag the respective commit as `v1.1.0` with a corresponding release.

docs/guide.md
```diff
@@ -142,22 +141,23 @@ and `<PATH_TO_INCLUDE_DIRECTORY>` is the path to the `include` directory of the
 
 ### Format for Comments
 
-For the information to be displayed correctly in the documentation, it is essential that the
-comments follow the format required by Doxygen. Below you find some tags, that are commonly used
-within the documentation of a function:
+For the information to be displayed correctly in the documentation, it the comments need to follow
```
denialhaag:
```suggestion
For the information to be displayed correctly in the documentation, the comments need to follow
```

docs/templates.md
```diff
@@ -31,32 +31,32 @@ device. The following sections describe how to set up and use the template.
...
 
-The default template is set up with the specified prefix and as a C++ project. You can easily switch
-the project to a purely C project by setting the `CXX_DEVICE` option to `OFF` in the
-`CMakeLists.txt` file in the root directory of the template.
+The default template is set up as a C++ project with the specified prefix. You can switch the
+project to a purely C project by setting the `CXX_DEVICE` option to `OFF` in the `CMakeLists.txt`
+file in the root directory of the template.
 
-For stability, we recommend to pin the version of QDMI that you are using for your implementation.
+For stability, we recommend pining the version of QDMI that you are using for your implementation.
```
denialhaag:
```suggestion
For stability, we recommend pinning the version of QDMI that you are using for your implementation.
```

docs/templates.md
```diff
@@ -65,10 +65,10 @@ variable in `cmake/ExternalDependecies.cmake` as follows:
 +   set(QDMI_REV "v1.0.0"
```
denialhaag: Should this reflect the most recent version, that is, v1.1.0?

examples/device/cxx/device.cpp
```diff
@@ -655,19 +651,196 @@ int CXX_QDMI_control_get_data_dev(CXX_QDMI_Job job,
...
+    return QDMI_ERROR_INVALIDARGUMENT;
+  }
+  if (session->status != CXX_QDMI_DEVICE_SESSION_STATUS::INITIALIZED) {
+    return QDMI_ERROR_BADSTATE;
+  }
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_NAME, "C++ Device with 5 qubits",
+                      prop, size, value, size_ret)
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_VERSION, "0.1.0", prop, size, value,
+                      size_ret)
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_LIBRARYVERSION, "1.0.0", prop, size,
```
denialhaag:
```suggestion
  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_LIBRARYVERSION, "1.1.0", prop, size,
```

denialhaag (commented):
Thanks for the quick implementation! The changes address our problems raised in #109. We look forward to updating our device when this PR is merged.

While looking at the docs and `device.cpp`, I found some minor typos. Otherwise, the changes look good to me.

docs/templates.md
burgholzer: Yeah. Thanks for caching this!
```suggestion
+   set(QDMI_REV "v1.1.0"
```

examples/device/c/device.c
```diff
@@ -408,243 +385,432 @@ int C_QDMI_control_check_dev(C_QDMI_Job job, QDMI_Job_Status *status) {
...
+    return QDMI_ERROR_INVALIDARGUMENT;
+  }
+  if (session->status != INITIALIZED) {
+    return QDMI_ERROR_BADSTATE;
+  }
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_NAME, "C Device with 5 qubits", prop,
+                      size, value, size_ret)
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_VERSION, "0.1.0", prop, size, value,
+                      size_ret)
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_LIBRARYVERSION, "1.0.0", prop, size,
```
burgholzer:
```suggestion
  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_LIBRARYVERSION, "1.1.0", prop, size,
```

burgholzer:
> Thanks for the quick implementation! The changes address our problems raised in #109. We look forward to updating our device when this PR is merged.
> 
> While looking at the docs and `device.cpp`, I found some minor typos. Otherwise, the changes look good to me.

Many thanks for catching these! Fixed in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/132/commits/188dc185c6b58cc35e513248e156c42dfda3960e ✅

mnfarooqi:
@burgholzer @ystade 
Nice work with addressing all the outstanding issues and organising the interface in one go.

Although I have an idea why the following might be there, it would be better if you could clarify the following: 
1) What's the reason for putting three enums (constants) in client.h when all the others are in `constants.h`? 
2) Why are there two job parameter enums `QDMI_JOB_PARAMETER_T` in `client.h` and `QDMI_DEVICE_JOB_PARAMETER_T` in `constants.h`?

burgholzer:
> @burgholzer @ystade Nice work with addressing all the outstanding issues and organising the interface in one go.
> 
> Although I have an idea why the following might be there, it would be better if you could clarify the following:
 
Thanks for raising these questions. I added two commits that hopefully address these to the rationale section of the documentation. You can preview them under the links below.

> 1. What's the reason for putting three enums (constants) in client.h when all the others are in `constants.h`?

https://munich-quantum-software-stack.github.io/QDMI/pr-preview/pr-132/md_docs_2rationale.html#rationale-job-structs

> 2. Why are there two job parameter enums `QDMI_JOB_PARAMETER_T` in `client.h` and `QDMI_DEVICE_JOB_PARAMETER_T` in `constants.h`?

https://munich-quantum-software-stack.github.io/QDMI/pr-preview/pr-132/md_docs_2rationale.html#rationale-enum-definitions

Feel free to suggest edits for any of the entries ☺️

echavarria-mqv:
Awesome work guys. I have no major concerns on my end. The PR looks good to go, please feel free to merge it.

burgholzer:
Alright. Thank you, everyone, for your feedback! Looks like this is good to go. 🚀 
I will kick off the release process tomorrow and, finally, close this first chapter in the history of QDMI 📖

PR #148, opened 2025-02-19
[WIP] Add workaround for manual doxygen build

### LINKED ISSUE #146: 🚸 Getting Doxygen via FetchContent
From #145: 
> As it turned out in #143, the fallback of getting Doxygen via FetchContent and building it from source when it is not installed in a sufficient version is not working in practice due to internal configuration and build errors in Doxygen (see https://github.com/doxygen/doxygen/issues/11416 for further details).
Thus, the documentation builds has been turned into an opt-in feature and, whenever enabled, now requires `Doxygen` via `find_package`.

We should revisit this whenever Doxygen resolves the respective issue to enable the docs build per default again and to not require developers to bother installing a sufficient version of Doxygen.

Gistbatch:
Closes: #146 

This is a preliminary workaround to get the manual installation of `doxygen` working.
I'm not sure if the line
```
target_compile_features(doxygen PUBLIC cxx_std_17)
```
is necessary; it seemed to have no effect, so I moved the c standard definition down.
Also, I chose to compare against `${DOXYGEN_EXECUTABLE}` as in one debugging session `${DOXYGEN_FOUND}` was true but the executable was not found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,14 +90,15 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS
     ON
     CACHE BOOL "Export compile commands" FORCE)
 
-# Require C standard
-set(CMAKE_C_STANDARD 11)
-set(CMAKE_C_STANDARD_REQUIRED True)
-set(CMAKE_C_EXTENSIONS False)
-
 include(cmake/ExternalDependencies.cmake)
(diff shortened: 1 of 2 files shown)


burgholzer:
This has, finally, been fixed in #250.
Thanks for pushing this at the beginning of the year!

PR #182, opened 2025-06-30
✨ Pulse representation specifications

### LINKED ISSUE #173: ✨ Data types for pulse representation

The following types with corresponding properties are proposed to be introduced:

`QDMI_PULSE_PARAMETER`:
- Name
- Range
- Permission (Read-only, R/W)

`QDMI_PULSE_SHAPE`: 
Predefined pulse shape:
- Name: e.g. “Gaussian”
- formula: str -> “ax^2+bx+c”
- parameters: List[`QDMI_PULSE_PARAMETER`]

`QDMI_PULSE_OPERATION_IMPLEMENTATION`:
- Pulse program Intermediate Representation (e.g., OpenPulse)
OR `QDMI_PULSE_SHAPE` with parameter values set

The types can likely be introduced as opaque pointers once the properties have been discussed and finalised.

mnfarooqi:
## Description

Introduce opaque pointers (`QDMI_Pulse_Parameter`, `QDMI_Pulse_Waveform` and `QDMI_Pulse_Implementation`), as well as the corresponding enums and query functions, to represent and query pulses.

Fixes #173 

## Checklist:

- [ ] The pull request only contains commits that are related to it.
- [ ] I have added appropriate tests and documentation.
- [ ] I have made sure all CI jobs on GitHub pass.
- [ ] The pull request introduces no new warnings and follows the project's style guidelines.


mnfarooqi:
Hi @donsano33 and @amitQC,
Could you please check and confirm whether this is what we need to represent a pulse?

examples/device/c/device.c
```diff
@@ -23,6 +23,8 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "c_qdmi/device.h"
 
+#include "c_qdmi/types.h"
+
```
ystade, line 27:
```suggestion
```
Iirc, this should not be necessary since the header above already re-exposes the types header.

examples/device/c/device.c
```diff
@@ -846,3 +848,29 @@ int C_QDMI_device_session_query_operation_property(
   }
   return QDMI_ERROR_NOTSUPPORTED;
 } /// [DOXYGEN FUNCTION END]
+
+struct C_QDMI_Pulse_Implementation_impl_d {};
+struct C_QDMI_Pulse_Waveform_impl_t {};
+struct C_QDMI_Pulse_Implementation_impl_t {};
```
ystade, line 854: The types, i.e., the structure names ending with `_t` should not needed to be redefined by the device, actually, this should lead to a compiler error because of redefinition. Am I overseeing something?

examples/device/c/device.c
```diff
@@ -846,3 +848,29 @@ int C_QDMI_device_session_query_operation_property(
...
+
+int C_QDMI_device_session_query_pulse_parameter_property(
+    C_QDMI_Device_Session session, C_QDMI_Pulse_Parameter param,
+    const QDMI_Pulse_Parameter_Property prop, const size_t size, void *value,
+    size_t *size_ret) {
+
```
ystade, line 860: Even if no property is supported, the function is expected to validate its arguments and return `QDMI_ERROR_INVALIDARGUMENT` if necessary. That has higher priority than `QDMI_ERROR_NOTSUPPORTED`.

ystade: This comment also applies to the two subsequent functions.

examples/device/cxx/device.cpp
```diff
@@ -24,6 +24,9 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "cxx_qdmi/device.h"
 
+#include "cxx_qdmi/types.h"
+#include "qdmi/constants.h"
```
ystade, line 28: Those should all be re-exported by the device header and not needed here. In particular, including headers from the directory with the prefixed headers and from the main include directory does not seem clean to me.
```suggestion
```

examples/device/cxx/device.cpp
```diff
@@ -212,7 +227,57 @@ const std::unordered_map<
           {{CXX_DEVICE_SITES[0], CXX_DEVICE_SITES[4]}, 0.95}}},
         // No need to specify single-qubit fidelities here
 };
-} // namespace
+
+const CXX_QDMI_Pulse_Parameter_impl_d AMPLITUDE{"A"};
```
ystade, line 231: If I understand that correctly, than only "A" is transferred as the name of the parameter through QDMI. Is that standardized anywhere? If not, I propose to use a more comprehensive name, i.e., `"amplitude"`. Similar reasoning applies to the next one.

examples/device/cxx/device.cpp
```diff
@@ -212,7 +227,57 @@ const std::unordered_map<
...
+    GAUSSIAN_PULSE_PARAMETERS = {&AMPLITUDE, &TIME, &SIGMA};
+
+const std::unordered_map<const CXX_QDMI_Pulse_Parameter_impl_d *,
+                         std::tuple<double, double, bool>>
+    GAUSSIAN_PULSE_PARAMETERS_PROPERTIES = {
+        {&AMPLITUDE, {0.3, 0.7, true}},
+        {&TIME,
+         {-std::numeric_limits<double>::infinity(),
+          std::numeric_limits<double>::infinity(), true}},
+        {&SIGMA, {10.0, 20.0, true}},
```
ystade, line 247: Even though QDMI does not define it, I propose to define a `struct` containing the three respective properties for a parameter instead of a tuple. This give the code better clearity as the different entries are than accessed via a meaningful name.

examples/device/cxx/device.cpp
```diff
@@ -776,7 +841,7 @@ int CXX_QDMI_device_session_query_device_property(
 
   ADD_SINGLE_VALUE_PROPERTY(
       QDMI_DEVICE_PROPERTY_PULSESUPPORT, QDMI_Device_Pulse_Support_Level,
-      QDMI_DEVICE_PULSE_SUPPORT_LEVEL_NONE, prop, size, value, size_ret)
+      QDMI_DEVICE_PULSE_SUPPORT_LEVEL_SITE, prop, size, value, size_ret)
```
ystade, line 844: Ok, that I a rather random place for this comment but this just came to my mind: To align this enum value to the naming convention throughout QDMI, I propose `QDMI_DEVICE_PULSESUPPORT_SITE`.

examples/device/cxx/device.cpp
```diff
@@ -873,3 +945,87 @@ int CXX_QDMI_device_session_query_operation_property(
...
+      (prop >= QDMI_PULSE_PARAMETER_PROPERTY_MAX &&
+       prop != QDMI_PULSE_PARAMETER_PROPERTY_CUSTOM1 &&
+       prop != QDMI_PULSE_PARAMETER_PROPERTY_CUSTOM2 &&
+       prop != QDMI_PULSE_PARAMETER_PROPERTY_CUSTOM3 &&
+       prop != QDMI_PULSE_PARAMETER_PROPERTY_CUSTOM4 &&
+       prop != QDMI_PULSE_PARAMETER_PROPERTY_CUSTOM5)) {
+    return QDMI_ERROR_INVALIDARGUMENT;
+  }
+
+  auto [range_min, range_max, is_mutable] =
```
ystade, line 963:
```suggestion
  const auto& [range_min, range_max, is_mutable] =
```
Saves some copies.

examples/device/cxx/device.cpp
```diff
@@ -873,3 +945,87 @@ int CXX_QDMI_device_session_query_operation_property(
...
+       prop != QDMI_PULSE_IMPLEMENTATION_PROPERTY_CUSTOM1 &&
+       prop != QDMI_PULSE_IMPLEMENTATION_PROPERTY_CUSTOM2 &&
+       prop != QDMI_PULSE_IMPLEMENTATION_PROPERTY_CUSTOM3 &&
+       prop != QDMI_PULSE_IMPLEMENTATION_PROPERTY_CUSTOM4 &&
+       prop != QDMI_PULSE_IMPLEMENTATION_PROPERTY_CUSTOM5)) {
+    return QDMI_ERROR_INVALIDARGUMENT;
+  }
+  ADD_SINGLE_VALUE_PROPERTY(
+      QDMI_PULSE_IMPLEMENTATION_PROPERTY_PULSEWAVEFORM, CXX_QDMI_Pulse_Waveform,
+      const_cast<CXX_QDMI_Pulse_Waveform>(
```
ystade, line 1021: I assume this `const_cast` is necessary, right? Just to double-check that it is not superfluous. Feel free to resolve immediately.

examples/driver/qdmi_example_driver.cpp
```diff
@@ -24,6 +24,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include "qdmi/client.h"
 #include "qdmi/device.h"
+#include "qdmi/types.h"
```
ystade, line 27:
```suggestion
```
Same reason as above.

test/test_qdmi.cpp
```diff
@@ -19,6 +19,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include "example_fomac.hpp"
 #include "example_tool.hpp"
 #include "qdmi/client.h"
+#include "qdmi/constants.h"
```
ystade, line 22:
```suggestion
```

test/utils/test_impl.cpp
```diff
@@ -36,6 +36,7 @@ void QDMIImplementationTest::SetUp() {
   const std::string &library_name = std::get<0>(params);
   const std::string &prefix = std::get<1>(params);
   mode = std::get<2>(params);
+  lib_prefix = std::get<1>(params);
```
ystade, line 39: As noted above, I would like to solve this differently and omit this string here.
```suggestion
```

test/utils/test_impl.hpp
```diff
@@ -54,6 +54,7 @@ class QDMIImplementationTest
   QDMI_Device device = nullptr;
   TEST_SESSION_MODE mode = TEST_SESSION_MODE::READONLY;
   std::string config_file_name;
+  std::string lib_prefix;
```
ystade, line 57: See the two comments above.
```suggestion
```

test/test_qdmi.cpp
```diff
@@ -972,5 +973,51 @@ TEST_P(QDMIImplementationTest, QueryPulseSupportLevel) {
...
   EXPECT_EQ(ret, QDMI_SUCCESS);
-  EXPECT_EQ(pulse_support_level, QDMI_DEVICE_PULSE_SUPPORT_LEVEL_NONE);
+  EXPECT_EQ(pulse_support_level, QDMI_DEVICE_PULSE_SUPPORT_LEVEL_SITE);
+}
+
+TEST_P(QDMIImplementationTest, QueryPulseImplementation) {
+
+  if (lib_prefix != "CXX") {
+    GTEST_SKIP() << "Skipping test for non-C++ implementations";
+  }
```
ystade, line 983: I find it contra-intuitive to decide which device offers a Pulse Implementation depending on the library name. There is an extra QDMI query function for that to query whether a device supports pulse level interaction. I see two possibilities:
1. Defining a macro, that can be added to every such test. This macro would query the pulse support and skip the test by calling `GTEST_SKIP` if necessary.
2. A class `QDMIPulseImplementationTest` that inherits from `QDMIImplementationTest` is defined. The `setUp` function checks the support for pulse level interaction and calls `GTEST_SKIP` if necessary. This class is then used by all test of the pulse level interface.

templates/device/test/test_my_device.cpp
```diff
@@ -17,6 +17,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ------------------------------------------------------------------------------*/
 
 #include "my_qdmi/device.h"
+#include "qdmi/constants.h"
```
ystade, line 20:
```suggestion
```

include/qdmi/device.h
```diff
@@ -351,7 +351,130 @@ int QDMI_device_session_query_operation_property(
...
+ * allow checking if the device supports the specified property without
+ * retrieving the property and without the need to provide a buffer for it.
+ * Additionally, the size of the buffer needed to retrieve the property is
+ * returned in @p size_ret if @p size_ret is not @c NULL.
+ *
+ * See the QDMI_device_query_pulse_waveform_property documentation for an
+ * example.
+ *
+ * May only be called after the session has been initialized with
+ * @ref QDMI_device_session_init.
```
ystade, line 431:
```suggestion
 * @remark Calling this function with @p value set to @c NULL is expected to
 * allow checking if the device supports the specified property without
 * retrieving the property and without the need to provide a buffer for it.
 * Additionally, the size of the buffer needed to retrieve the property is
 * returned in @p size_ret if @p size_ret is not @c NULL.
 * See the QDMI_device_query_pulse_waveform_property documentation for an
 * example.
 * @attention May only be called after the session has been initialized with
 * @ref QDMI_device_session_init.
```
to make it in line with the docstring above.

include/qdmi/constants.h
```diff
@@ -453,6 +453,9 @@ enum QDMI_OPERATION_PROPERTY_T {
   QDMI_OPERATION_PROPERTY_DURATION = 3,
   /// `double` The fidelity of an operation.
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
+  /// @ref QDMI_Pulse_Implementation The pulse implementation of the
+  /// operation.
+  QDMI_OPERATION_PROPERTY_PULSEIMPLEMENTATION = 5,
```
ystade, line 458: Why does it actually need a `QDMI_Pulse_Implementation` entity? If I understand correctly than the mapping of operations to pulse implementations is a one-to-one mapping. Hence, keeping the interface minimal and because I do not see a significant difference between operation and pulse implementation, I propose to remove this entity and incorporate all of its properties that can be queried as operation properties. A device may report `NOTSUPPORTED` if it does not support pulse level control.

include/qdmi/constants.h
```diff
@@ -482,6 +485,180 @@ enum QDMI_OPERATION_PROPERTY_T {
 /// Operation property type.
 typedef enum QDMI_OPERATION_PROPERTY_T QDMI_Operation_Property;
 
+/**
+ * @brief Enum of the pulse parameter properties that can be queried via @ref
+ * QDMI_device_session_query_pulse_parameter_property as part of the @ref
+ * device_interface "device interface" and via @ref
+ * QDMI_device_query_pulse_parameter_property as part of the @ref
+ * client_interface "client interface".
```
ystade, line 493: I hope we did not vialoate this in the rest of QDMI, but a `@brief` description is a one-liner. You can move additional information into `@details`.

include/qdmi/constants.h
```diff
@@ -482,6 +485,180 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+   * parameter can take.
+   */
+  QDMI_PULSE_PARAMETER_PROPERTY_RANGEMAX = 2,
+  /**
+   * @brief `bool` Whether the pulse parameter is mutable.
+   * @details If this property is set to true, the pulse parameter can be
+   * modified by the user. If it is set to false, the pulse parameter is fixed
+   * and cannot be changed.
+   */
+  QDMI_PULSE_PARAMETER_PROPERTY_MUTABLE = 3,
```
ystade, line 523: I do not understand the purpose of immutable parameters, those are constants. I suggest that constants can already occur as constants in the formula.

include/qdmi/constants.h
```diff
@@ -482,6 +485,180 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+
+/// Pulse parameter property type.
+typedef enum QDMI_PULSE_PARAMETER_PROPERTY_T QDMI_Pulse_Parameter_Property;
+
+/**
+ * @brief Enum of the pulse waveform properties that can be queried via @ref
+ * QDMI_device_session_query_pulse_waveform_property as part of the @ref
+ * device_interface "device interface" and via @ref
+ * QDMI_device_query_pulse_waveform_property as part of the @ref
+ * client_interface "client interface".
```
ystade, line 555: Same as above, should be a one-liner.

include/qdmi/constants.h
```diff
@@ -482,6 +485,180 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+ */
+enum QDMI_PULSE_WAVEFORM_PROPERTY_T {
+  /**
+   * @brief `char*` (string) The name is used to identify the pulse waveform.
+   */
+  QDMI_PULSE_WAVEFORM_PROPERTY_NAME = 0,
+  /**
+   * @brief `char*` (string) The formula used to generate the pulse waveform.
+   */
+  QDMI_PULSE_WAVEFORM_PROPERTY_FORMULA = 1,
```
ystade, line 568: Ok, this will be a bigger comment. I'll put that in my general comment such that we can discuss and consolidate this in the main thread of this PR.

include/qdmi/constants.h
```diff
@@ -482,6 +485,180 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+ * queried to query the pulse program in the desired or in the available format.
+ */
+enum QDMI_PULSE_IMPLEMENTATION_PROPERTY_T {
+  /**
+   * @brief `char*` (string) The pulse implementation in the Open Pulse format.
+   */
+  QDMI_PULSE_IMPLEMENTATION_PROPERTY_OPENPULSE = 0,
+  /**
+   * @brief `QDMI_Pulse_Waveform` (@ref QDMI_Pulse_Waveform) The pulse
+   * implementation as pulse waveform.
```
ystade, line 620:
```suggestion
   * implementation as a pulse waveform.
```
?

ystade (changes requested):
Thanks @mnfarooqi for this first proposal for the pulse level support. Your implementation is already very clean and follows the look and feel of QDMI.

I went though it skipping the FoMaC implementation for now because I think, we should first dicuss some high level topics first. In a later iteration, we can also consolidate FoMaC and the testing.

As indicated in one of the comments, I have one major point of concern: With `QDMI_Pulse_Waveform` we are defining our own standard for representing pulses. Honestly, I am not an expert in pulse level quantum computing. However, I strongly argue that introducing a new calibration grammar should be well-considered. This concern accumulates in the following two points:
1. The current proposal allows to query the waveform in the OpenPulse format. Why do I then even need to have another representation as OpenPulse seems to be supported. I suggest that we decide to use one representation and not multiple to keep the interface minimalistic and do not impose additional implemenation efforts on device implementers.
2. The current (QDMI) waveform specification relies on an unspecified string. First of all, I would argue that a string to represent a term is not the best choice. However, what I find more important to fix the semantics of the pulse representation. Right now, a compiler cannot expect anything from the returned string. For example, consider, the [OpenPulse Grammar](https://openqasm.com/language/openpulse.html). If we do not want to rely on open pulse (for whatever reason), we need a semantic for our pulse representation as well.

@mnfarooqi sorry, questioning the pulse representation. This has nothing to do with the actual code. Your code is perfectly well-strucutred, well-documented, and clean. Nevertheless, we should consolidate what we actually want here.

mnfarooqi:
Thank you @ystade for the detailed review and for moving the discussion forward. The current implementation is based on our initial discussion and is not final, but is intended to facilitate the discussion. I will address the inline comments once we have agreed on the basics. I am also not an expert on pulses, which is why I tagged @donsano33 and @amitQC. Hopefully, they will add their review soon.

The pulse representation here is supposed to serve two purposes: one is to query the generic pulse shapes supported by the device, along with the constraints and pulses for specific operations. The second purpose is to set or overwrite existing pulses for a specific operation.

> 
> 1. The current proposal allows to query the waveform in the OpenPulse format. Why do I then even need to have another representation as OpenPulse seems to be supported. I suggest that we decide to use one representation and not multiple to keep the interface minimalistic and do not impose additional implemenation efforts on device implementers.

The OpenPulse Grammar can be used to query and set pulses specific to an operation, but as far as I know, it cannot be used to query generic pulse waveforms, i.e. the pulse formula and the limits of the parameters in that formula.

> 2. The current (QDMI) waveform specification relies on an unspecified string. First of all, I would argue that a string to represent a term is not the best choice. However, what I find more important to fix the semantics of the pulse representation. Right now, a compiler cannot expect anything from the returned string. For example, consider, the [OpenPulse Grammar](https://openqasm.com/language/openpulse.html). If we do not want to rely on open pulse (for whatever reason), we need a semantic for our pulse representation as well.
> 

As far as I understand, the queried generic pulse waveform is to be consumed manually by a human, not by a compiler. The only automation tool I can think of at the moment would be one that consumes this information and generates human-readable documentation on the pulses supported by a device, along with the parameters and the limits on those parameters. I searched for a standard notation that could be specified in the QDMI specification, but I could not find one. I am thinking of proposing infix notation, which could be used by a tool such as [str2sym](https://de.mathworks.com/help/symbolic/str2sym.html) from MATLAB. If you have any other ideas, please share them.

amitQC:
Thanks @mnfarooqi and @ystade  for starting the conversation. The idea of `waveform`, or in other words, `envelope`, is to query the existing waveforms, e.g., Gaussian, sine, etc. (see here: https://openqasm.com/language/openpulse.html#waveforms). These are generally an array of complex values. We also wanted to provide a mathematical description of these waveforms or envelopes, where a user can change the parameter values of an envelope to generate a pulse. For example, a `sine` envelope is of the form: $\Sigma_{n} a_{n} sin(n *pi *t/T)$, where one just needs to define values of $a_{n}$ to define a pulse. 

From the compiler's perspective, in my opinion, the only thing that matters is the corresponding operation to a pulse. For example, if a NOT gate has a sine waveform, you can take the performance of this pulse into account in FoMaC, but not the underlying pulse itself.

ystade:
Would you mind explaining the following point in your answer a bit further? I do not think that I understand this correctly. What information exactly matters to a compiler? How do I get from a random (I only write random because right now the string is not further specified and could in principle be anything) string to the _performance_?

> you can take the performance of this pulse into account in FoMaC, but not the underlying pulse itself.

amitQC:
> Would you mind explaining the following point in your answer a bit further? I do not think that I understand this correctly. What information exactly matters to a compiler? How do I get from a random (I only write random because right now the string is not further specified and could in principle be anything) string to the _performance_?
> 
> > you can take the performance of this pulse into account in FoMaC, but not the underlying pulse itself.

Sure! Depending on the level of compilation, for a top-level, the underlying pulse implementing a gate `U` does not matter; however, one can consider pulse performance, such as fidelity, time, etc., into consideration. I think that there should be a performance metric corresponding to each gate (i.e., underlying pulse) that can be queried by the compiler. For a very low-level compiler, one can also consider compiling only at the level of pulses, i.e., convert every gate `U` object into pulses and then try compilation. This might help to further compress the resulting circuit (however, there are no general guarantees; this is hardware-specific). Does this answer your question?

mnfarooqi:
> > Would you mind explaining the following point in your answer a bit further? I do not think that I understand this correctly. What information exactly matters to a compiler? How do I get from a random (I only write random because right now the string is not further specified and could in principle be anything) string to the _performance_?
> > > you can take the performance of this pulse into account in FoMaC, but not the underlying pulse itself.
> 
> Sure! Depending on the level of compilation, for a top-level, the underlying pulse implementing a gate `U` does not matter; however, one can consider pulse performance, such as fidelity, time, etc., into consideration. I think that there should be a performance metric corresponding to each gate (i.e., underlying pulse) that can be queried by the compiler. For a very low-level compiler, one can also consider compiling only at the level of pulses, i.e., convert every gate `U` object into pulses and then try compilation. This might help to further compress the resulting circuit (however, there are no general guarantees; this is hardware-specific). Does this answer your question?

Does that mean that the compiler will not use the pulse shape formula (i.e. the mathematical description of these waveforms or envelopes) directly? If not, where/who can use it?

amitQC:
> > > Would you mind explaining the following point in your answer a bit further? I do not think that I understand this correctly. What information exactly matters to a compiler? How do I get from a random (I only write random because right now the string is not further specified and could in principle be anything) string to the _performance_?
> > > > you can take the performance of this pulse into account in FoMaC, but not the underlying pulse itself.
> > 
> > 
> > Sure! Depending on the level of compilation, for a top-level, the underlying pulse implementing a gate `U` does not matter; however, one can consider pulse performance, such as fidelity, time, etc., into consideration. I think that there should be a performance metric corresponding to each gate (i.e., underlying pulse) that can be queried by the compiler. For a very low-level compiler, one can also consider compiling only at the level of pulses, i.e., convert every gate `U` object into pulses and then try compilation. This might help to further compress the resulting circuit (however, there are no general guarantees; this is hardware-specific). Does this answer your question?
> 
> Does that mean that the compiler will not use the pulse shape formula (i.e. the mathematical description of these waveforms or envelopes) directly? If not, where/who can use it?

If compilers use the mathematical formula, it might be a very complex process (as the number of gates and qubits increases), because you have to deal with some complex parameters, compared to an object `U`. These formulas can be used by the user to define the pulse corresponding to a gate `U`. For example, see waveforms defined here: https://docs.meetiqm.com/iqm-pulse/api/iqm.pulse.playlist.waveforms.html.

In my opinion (and as we also briefly discussed during our meetings), one can design pulses corresponding to beyond what the hardware offers, i.e., native gate sets using the formulas mentioned above or by giving a full complex array of amplitudes. If these are different pulse implementations corresponding to a gate `U` (say, `P1`, `P2`, `P3`, etc.), and each pulse `P` has a different property, i.e., duration, fidelity, noise-robustness, etc. These various implementations of `U` can be used by the compiler. For example, if the circuit is deep, the compiler should choose the pulse with the shortest duration.

mnfarooqi:
Moving this thread forward, do either of you have any major comments or suggestions before I address the detailed comments on the code, @burgholzer and @ystade?

burgholzer:
> Moving this thread forward, do either of you have any major comments or suggestions before I address the detailed comments on the code, @burgholzer and @ystade?

I think the most important high level comment is still the one from Yannick (https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/182#pullrequestreview-2991318535)
As long as you keep that in mind while working on the code, that should work.
Note that Yannick is still on vacation the next week. I'll try to answer any questions that come up.

burgholzer:
I am moving this to the 1.3.0 milestone for now as there still seems to be ongoing discussions that need to be resolved and the changes on `main` have been pretty stable for a while.

burgholzer:
I am closing this PR for now while keeping its branch alive for posterity.
This is unlikely to land in QDMI v1 if no one pulls of a heroic effort and is much better suited as a dedicated (sub)module in QDMI v2.

PR #186, opened 2025-07-03
🐛 Fix Device Template

### LINKED ISSUE #169: 🐛 Leftover after Removal of Device Sites and Operations
As far as I know, the type for `QDMI_Device_Site`s and `QDMI_Device_Operation`s was removed for a good reason. I think this change did not make it into the template. What surprises me is that this is not detected by the test CI pipeline as the actual required types are incomplete then.

For reference:

https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/3aec97c03c714584c498cbe6b3acff7d0df1bd79/templates/device/src/my_device.cpp#L56-L66

To fix this:
1. One must remove `Device_` from those types in the above snippet
2. One could investigate why this error was not caught by the CI and try to find a solution for it.

ystade:
## Description

There was a leftover in the device template from older QDMI versions in the definition of sites and operations as mentioned in #169 . This is fixed by this PR.

Fixes #169 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ clients compiled against a different minor or major version.
 
 ### Added
 
+- 🔥 Remove example device implementation in C ([#169]) ([\@ystade])
 - 📝 Add changelog and upgrade guide ([#160]) ([\@ystade], [\@burgholzer])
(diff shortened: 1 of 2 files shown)


burgholzer (approved):
LGTM!
Worth a changelog entry?

ystade:
Why, not. I'll add one and merge afterwards.

PR #198, opened 2025-08-05
✨ Communicating the Location of Sites

### LINKED ISSUE #185: ✨ Communicating the Location of Sites

As pointed out in the parent issue #184, the location of the atom, i.e., qubits, plays a crucial role in their possible interaction with other atoms. Hence, a compiler must be aware of the possible locations an atom can assume. One central part of (so far all existing) neutral atom devices is static traps created by optical tweezers or an optical lattice. In any case, there is a set of sites (`QDMI_Site`) that can hold up to one atom, i.e., it could also be empty. Important to the compiler is the location of those sites.

The straightforward solution to this problem is to add two further enumeration values to
https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/4c517b9dd85a2a366d10b9d498565f6705e9f3c0/include/qdmi/constants.h#L389

Namely,
```c++
  /**
   * @brief `double` The x-coordinate of the site.
   * @details The x-coordinate is measured relative to some unique origin of the device, i.e., the pair 
   * of x- and y-coordinate (@ref QDMI_SITE_PROPERTY_YCOORDINATE) must be unique to the site.
   * @note This property is mainly required for neutral atom devices to report the location of sites.
   */
  QDMI_SITE_PROPERTY_XCOORDINATE,
  
  /**
   * @brief `double` The y-coordinate of the site.
   * @details The y-coordinate is measured relative to some unique origin of the device, i.e., the pair 
   * of x- (@ref QDMI_SITE_PROPERTY_XCOORDINATE) and y-coordinate must be unique to the site.
   * @note This property is mainly required for neutral atom devices to report the location of sites.
   */
  QDMI_SITE_PROPERTY_YCOORDINATE,
```

There remains an open question: Should QDMI fix the unit of those coordinates, i.e., define that those coordinates are measured in micrometers (µm), or should there be a way for the device to choose its unit and communicate this through QDMI? This could be realized as a pair of two
https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/4c517b9dd85a2a366d10b9d498565f6705e9f3c0/include/qdmi/constants.h#L277
returning
1. a unit as a string, e.g., `"um"` together with
2. a factor all length units (including these coordinates) must be multiplied with to get the actual length in the above unit.

This could then look as follows:
```c++
  /**
   * @brief `char*` (string) The length unit used by the device.
   * @details This property must be a known SI unit, e.g., "mm", "um" or "nm". All length values are first
   * multiplied with the value returned by @ref QDMI_DEVICE_PROPERTY_UNITVALUE and then interpreted
   * in this unit.
   * @note If the device returns any length value, this property must be set.
   */
  QDMI_DEVICE_PROPERTY_UNIT,
  
  /**
   * @brief `double` A factor applied to all length values.
   * @details This value must be multiplied with all length values returned by the device before it can be
   * interpreted in the unit returned by  @refQDMI_DEVICE_PROPERTY_UNIT.
   * @note If this property is @ref QDMI_ERROR_NOTSUPPORTED, a default value of `1.0` is assumed.
   */
  QDMI_DEVICE_PROPERTY_UNITVALUE,
```

ystade:
## Description

This PR implements the features requested in #185.

Fixes #185

## Checklist:

### DIFF
--- a/examples/device/device.cpp
+++ b/examples/device/device.cpp
@@ -31,6 +31,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include <cstdint>
 #include <cstring>
 #include <functional>
+#include <iostream>
 #include <iterator>
 #include <limits>
 #include <map>
@@ -776,6 +777,11 @@ int CXX_QDMI_device_session_query_device_property(
   ADD_SINGLE_VALUE_PROPERTY(
       QDMI_DEVICE_PROPERTY_PULSESUPPORT, QDMI_Device_Pulse_Support_Level,
       QDMI_DEVICE_PULSE_SUPPORT_LEVEL_NONE, prop, size, value, size_ret)
+
+  ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_LENGTHUNIT, "um", prop, size, value,
+                      size_ret)
+  ADD_SINGLE_VALUE_PROPERTY(QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR, double, 1.0,
+                            prop, size, value, size_ret)
 
   return QDMI_ERROR_NOTSUPPORTED;
 } /// [DOXYGEN FUNCTION END]
--- a/examples/device/device.cpp
+++ b/examples/device/device.cpp
@@ -31,7 +31,6 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include <cstdint>
 #include <cstring>
 #include <functional>
-#include <iostream>
 #include <iterator>
 #include <limits>
 #include <map>
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -48,7 +48,7 @@ enum QDMI_STATUS {
   QDMI_ERROR_NOTSUPPORTED = -9,     ///< Operation is not supported.
   /// Resource is in the wrong state for the operation.
   QDMI_ERROR_BADSTATE = -10,
-  QDMI_ERROR_TIMEOUT = -11 ///< Operation timed out.
+  QDMI_ERROR_TIMEOUT = -11, ///< Operation timed out.
 };
 
 /**
@@ -334,8 +334,9 @@ enum QDMI_DEVICE_PROPERTY_T {
   /**
    * @brief `char*` (string) The length unit used by the device.
    * @details This property must be a known SI unit, e.g., "mm", "um" or "nm".
-   * All length values are first multiplied with the value returned by @ref
-   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR and then interpreted in this unit.
+   * All length values must first be multiplied with the value returned by @ref
+   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR and then interpreted in the unit
+   * set for this property here.
    * @note If the device returns any length value, this property must be set.
    */
   QDMI_DEVICE_PROPERTY_LENGTHUNIT = 10,
@@ -426,7 +427,7 @@ enum QDMI_SITE_PROPERTY_T {
    */
   QDMI_SITE_PROPERTY_NAME = 3,
   /**
-   * @brief `double` The x-coordinate of the site.
+   * @brief `int64_t` The x-coordinate of the site.
    * @details The x-coordinate is measured relative to some unique origin of the
    * device, i.e., the triple of x-, y- (@ref QDMI_SITE_PROPERTY_YCOORDINATE),
    * and z-coordinate (@ref QDMI_SITE_PROPERTY_ZCOORDINATE) must be unique to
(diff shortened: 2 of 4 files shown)


include/qdmi/constants.h
```diff
@@ -48,7 +48,7 @@ enum QDMI_STATUS {
   QDMI_ERROR_NOTSUPPORTED = -9,     ///< Operation is not supported.
   /// Resource is in the wrong state for the operation.
   QDMI_ERROR_BADSTATE = -10,
-  QDMI_ERROR_TIMEOUT = -11, ///< Operation timed out.
+  QDMI_ERROR_TIMEOUT = -11 ///< Operation timed out.
```
```suggestion
  QDMI_ERROR_TIMEOUT = -11, ///< Operation timed out.
```
Seems like an unintentional change. I'd rather keep the comma as this way adding a new enum member does not cause changes in unrelated lines.

include/qdmi/constants.h
```diff
@@ -408,6 +425,48 @@ enum QDMI_SITE_PROPERTY_T {
...
+   * (@ref QDMI_SITE_PROPERTY_YCOORDINATE), and z-coordinate must be unique to
+   * the site.
+   * @note This property is mainly required for neutral atom devices to report
+   * the location of sites.
+   * @note This property is a length value and must be interpreted in the way
+   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
+   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
+   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
+   */
+  QDMI_SITE_PROPERTY_ZCOORDINATE = 6,
```
burgholzer: Should these really be `double` values and not `int64_t`?
We briefly discussed about this at some point, but I don't quite remember the conclusion.
I thought that a discretization of 3D space was what we ended up with as a solution going forward.

examples/device/device.cpp
burgholzer, line 1: Wouldn't it be nice to also show the coordinates in action here and test that the scaling factor works appropriately, i.e., have it be a non-1.0 factor and determine the location of certain sites checking that the computation is conducted correctly?

The position for Site `i` could simply be `(i, i, i)`.

burgholzer (changes requested):
This is looking good 👍🏼 I only have three small little remarks/comments/questions before this can go in.

include/qdmi/constants.h
ystade: I would happily agree if that would not be a C99 feature (and is prohibited in strict C98 and hence causes a compiler warning) and all other enums do not do that. What do you say?

include/qdmi/constants.h
ystade: Yes, for the representation itself, we concluded using integers. However, for the interface double seems more flexible, does it?
For context: The device would use ints internally and how the client then interprets the coordinates, directly as double or casting them back to ints is up to the client.

include/qdmi/constants.h
burgholzer: We use a way more modern version of C than all of that. So it is definitely allowed. If this is throwing a warning, we are not setting a flag correctly somewhere that sets the C standard.

examples/device/device.cpp
ystade: What do you mean with "the computation is conducted correctly"? The device would for each related property just return a predefined fixed constant. The computation you are referring to only happens in the client.

include/qdmi/constants.h
burgholzer: I don't particularly know if flexibility is an advantage here to be honest.
But, on the other hand, I do not have strong feelings on this.
As long as using `ints` is an option (which is it when this is expected to be a `double`) and we are using integers in everything we do, I am fine.

include/qdmi/constants.h
ystade: Ok, then I saw the issue in some other project, where I included QDMI. Then it is probably the problem of the setup in the other project.

examples/device/device.cpp
burgholzer: Got you. In that case the following wording is misleading:
> * All length values are first multiplied with the value returned by @ref
> * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR and then interpreted in this unit.
   
This makes it sound like the device is doing that when retrieving length values.
The `QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR` docstring is slightly clearer in that regard.

include/qdmi/constants.h
ystade: I am also open to using ints in the interface itself until someone comes around and complains why we are not using double here.

burgholzer: Then they can use the scaling factor, which is a `double` to regulate that ;o)
I'd prefer the `int64_t` solution here.

ystade:
I guess the last commit addresses all raised concerns, @burgholzer .

burgholzer (approved):
Yeah. LGTM 👍🏼 feel free to merge once CI is green

Issue #184, opened 2025-07-03
✨ Neutral Atom Support


ystade:
### What's the problem this feature will solve?

Every technology comes with its specific properties and features, and so do neutral atom devices. For neutral atom devices, qubits are represented by neutral atoms that are trapped in optical tweezers. These tweezers can be static, i.e., their location is fixed, or they can be dynamically rearranged. Additionally, it is not predetermined which qubits can interact with each other, and no coupling map can be queried. Instead, the interaction of atoms depends on their current distance. Note that due to dynamic rearrangements, the distances and hence the connectivity of qubits can change during computation.

### Describe the solution you'd like

To properly reflect the properties of neutral atom devices, other information than for superconducting devices must be transferred through QDMI, such that, e.g., a compiler has the required information available to compile circuits to target-specific instructions for a neutral atom device. The associated sub-issues will discuss individual properties specific to neutral atom devices. The aim of this main issue, together with all sub-issues, is to make QDMI ready such that all relevant information of the neutral atom device, e.g., for the compiler, can be transferred through the interface.

burgholzer:
All the sub-issues here have been resolved and I think it is fair to say that we can claim initial support for neutral atom platforms here.
As we are preparing for the release of 1.2, we can close this chapter for now and continue it as part of the next minor release cycle.

Issue #189, opened 2025-07-03
✨ Representing Neutral Atom Arrays


ystade:
### What's the problem this feature will solve?

Neutral atom devices usually feature one or multiple arrays of traps, i.e., sites where atoms can be located. On both sides, i.e., in devices and clients, it is tedious to describe those arrays of sites that usually exhibit a regular structure, just as a list of sites. However, currently, that is the only way information about sites can be queried through QDMI. The sites are returned just as a list. In this regard, #185 aims to provide the x- and y-coordinates of each site. However, that does not reveal any information about the topology, i.e., which sites form an array or a group.

### Describe the solution you'd like

One way to efficiently describe neutral atom devices with multiple zones would be to define the following properties for every array of sites:
- `origin`: The origin of the array as 2D coordinates
- `sublattice`: This is a set of 2D coordinates that defines sites relative to a reference point.
- `vector1`, `vector2`: Two 2D-basis vectors. The sites defined by the sublattice above are copied by multiples of the two basis vectors.
- `extent`: Definition of the extent of the array. Sites defined by the method above only exist within the defined extent, which is most likely a rectangle.

**Note:** The sublattice allows to define more complicated structures than simple grids and skew grids such as honeycomb structures.

While it is not necessary to transfer all this information for every atom array through QDMI, some extra information is required to efficiently reconstruct exactly the information used by the device on the client side. This would be:
- `QDMI_SITE_PROPERTY_MODULEINDEX`: an integer that uniquely identifies the atom array.
- `QDMI_SITE_PROPERTY_SUBMODULEINDEX`: an integer uniquely identifying the sublattice within a module.

Analogously, for the in #188 mentioned Rigetti topology, the individual rings could be uniquely identified by the sub-module index. If multiple modules are connected via inter-chip connects, each chip could be uniquely identified by the module index.

burgholzer:
Just to spell that out: This would either require a new opaque data type for the array of sites that provides all that information (origin, sublattice, etc.) or new `QDMI_DEVICE_PROPERTY_...` that allow querying this topology information from the device, right?

The `ModuleIndex` and `SubModuleIndex` make sense, and would just be new, optional `QDMI_SITE_PROPERTY` members.

Just a fun fact:
[...]

PR #199, opened 2025-08-05
✨ Report the Scope of an Operation

ystade:
## Description

On NA devices, operations on qubits can be global, i.e., a laser illuminates all atoms within an array, inducing an operation on all illuminated atoms. For compilation, it is important to know, whether an operation can only be applied to all atoms simultaneously or whether it can be addressed to individual qubits.

To this end, the concept of sites is generalised. The sites used previously remain sites and are called "regular sites". Additionally, to those this PR introduces "zone sites". Those sites report being a zone by a corresponding property. Additionally they define a spatial extent.

A global operation acting on a zone represented by a `QDMI_Site` object `interaction_zone` would then report `QDMI_ERROR_UNSUPPORTED` for all other sites if queried through the function `QDMI_device_query_operation_property`.

In conclusion, determining whether an operation is global on some zone is a two-fold process:
1. One finds the site(s) and operation can be applied to.
2. If the found site(s) are zones, i.e., the property `QDMI_SITE_PROPERTY_ISZONE` is `true`, then the operation is global.

Since this PR is concerned with NA-specific support of QDMI, it also contributes to #184.

## Checklist:

### DIFF
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -227,3 +227,22 @@ the same enumeration, each with a different prefix. It would also not be possibl
 know about all the different values of the enumeration. Additionally, when linking all devices
 statically into the driver, the name-shifted header for each device must be included. If each device
 would define the enumerations anew, this would lead to a compile-time error.
+
+## Why are there different kinds of sites? {#rationale-site-types}
+
+Originally, QDMI only defined a single kind of site, which was used to represent a location that can
+potentially hold a qubit. Already from the start, we intentionally avoided using the term `Qubit` to
+refer to their location, as on some devices, e.g., neutral atom-based ones, a site must not
+necessarily hold an atom representing the qubit. Hence, a site is more general than a qubit and can
+represent any kind of location that can hold a qubit, such as a superconducting qubit, a neutral
+atom, or a trapped ion. However, when extending the QDMI support for neutral atom-based devices, we
+realized that this was not general enough. These devices offer _global_ operations that execute on
+multiple atoms within a given zone simultaneously. In particular, these global operations cannot be
+executed on individual atoms but on all atoms within a zone at once. Consequently, we had to equip
+QDMI such that it is capable of reporting operation properties with respect to zones and not only
+individual qubit positions. To this end, we introduced the concept of _zone sites_ that represent a
+spatial zone. These have the same type as regular sites, i.e., @ref QDMI_Site, but are used to
+represent a zone instead of a single qubit position. Hence, they can be used in the function @ref
+QDMI_device_query_operation_property as parameter to query data like fidelity or duration of global
+operations specific to a zone. This allows QDMI to represent the capabilities of neutral atom-based
+devices while keeping the interface consistent with the existing site concept.
--- a/docs/rationale.md
+++ b/docs/rationale.md
@@ -244,5 +244,7 @@ individual qubit positions. To this end, we introduced the concept of _zone site
 spatial zone. These have the same type as regular sites, i.e., @ref QDMI_Site, but are used to
 represent a zone instead of a single qubit position. Hence, they can be used in the function @ref
 QDMI_device_query_operation_property as parameter to query data like fidelity or duration of global
-operations specific to a zone. This allows QDMI to represent the capabilities of neutral atom-based
-devices while keeping the interface consistent with the existing site concept.
+operations specific to a zone. The function @ref QDMI_device_query_device_property will return a
+list of all, i.e., regular and zone sites, for the property @ref QDMI_DEVICE_PROPERTY_SITES.
+Overall, these modifications allow QDMI to represent the capabilities of neutral atom-based devices
+while keeping the interface consistent with the existing site concept.
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -514,14 +514,12 @@ enum QDMI_OPERATION_PROPERTY_T {
   /// `double` The fidelity of an operation.
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
   /**
-   * @brief @ref QDMI_Operation_Scope The scope of the operation.
-   * @details The scope of an operation indicates whether the operation
-   * executes on individual sites or on all sites this operation can be applied
-   * to.
-   * @note This property is mainly required for neutral atom devices that offer
-   * global operations in addition to local operations.
+   * @brief `bool` An operation is global if it acts on all sites
+   * simultaneously. the operation is applicable to.
+   * @note This property is mainly relevant for neutral atom devices that can
+   * perform operations on all sites at once, e.g., by global laser pulses.
    */
-  QDMI_OPERATION_PROPERTY_SCOPE = 5,
+  QDMI_OPERATION_PROPERTY_GLOBAL = 5,
   /**
    * @brief The maximum value of the enum.
    * @details It can be used by devices for bounds checking and validation of
@@ -875,26 +873,6 @@ enum QDMI_DEVICE_PULSE_SUPPORT_LEVEL_T {
 
 /// Pulse support level type.
 typedef enum QDMI_DEVICE_PULSE_SUPPORT_LEVEL_T QDMI_Device_Pulse_Support_Level;
-
-/// The scope of an operation, e.g., local on a site or global on all sites.
-enum QDMI_OPERATION_SCOPE_T {
-  /**
-   * @brief The operation is local, i.e., it acts on individual sites.
-   * @details If the gate is a n-qubit gate, it acts exactly on the n sites
-   * that are specified in the program.
-   */
-  QDMI_OPERATION_SCOPE_LOCAL = 0,
-  /**
-   * @brief The operation is global, i.e., it acts on all sites the operation is
-   * applicable to.
-   * @note This scope is mainly used for neutral atom devices that can
-   * perform operations on all sites at once, e.g., global laser pulses.
-   */
-  QDMI_OPERATION_SCOPE_GLOBAL = 1
-};
-
-/// Operation's scope type.
-typedef enum QDMI_OPERATION_SCOPE_T QDMI_Operation_Scope;
 
 // NOLINTEND(performance-enum-size, modernize-use-using)
 
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -513,6 +513,15 @@ enum QDMI_OPERATION_PROPERTY_T {
   /// `double` The fidelity of an operation.
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
   /**
+   * @brief @ref QDMI_Operation_Scope The scope of the operation.
+   * @details The scope of an operation indicates whether the operation
+   * executes on individual sites or on all sites within the region of
+   * the operation.
+   * @note This property is mainly required for neutral atom devices that offer
+   * apart from local operations also global operations.
+   */
+  QDMI_OPERATION_PROPERTY_SCOPE = 5,
+  /**
    * @brief The maximum value of the enum.
    * @details It can be used by devices for bounds checking and validation of
    * function parameters.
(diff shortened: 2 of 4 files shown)


ystade:
~~⚠️ Note, this PR is currently based on the changes of #189 and must be rebased on `main` before merging.~~

include/qdmi/constants.h
```diff
@@ -860,12 +869,32 @@ enum QDMI_DEVICE_PULSE_SUPPORT_LEVEL_T {
...
+   * @brief The operation is global, i.e., it acts on all sites the operation is
+   * applicable to.
+   * @note This scope is mainly used for neutral atom devices that can
+   * perform operations on all sites at once, e.g., global laser pulses.
+   */
+  QDMI_OPERATION_SCOPE_GLOBAL = 1
+};
+
+/// Operation's scope type.
+typedef enum QDMI_OPERATION_SCOPE_T QDMI_Operation_Scope;
```
burgholzer: Do we foresee that this will, at some point, contain more than two members? Otherwise, this would not need an explicit enumeration, but could simply be a Boolean flag indicating whether the operation is local or global.

Follow-up question: How does this work for an operation in a zone-based NA architecture with multiple processing zones?
The corresponding CZ operation would be global to the zone, but not global on all sites.
Right now, we can query QDMI for the sites an operation is supported on (right?). However, that would yield a 1D list of sites at the moment. This would not let us distinguish between different zones.
I am probably overlooking something here, so please correct me if I am wrong.

include/qdmi/constants.h
```diff
@@ -860,12 +869,32 @@ enum QDMI_DEVICE_PULSE_SUPPORT_LEVEL_T {
    * @details This means that the device can execute pulse-level instructions on
    * both the sites and channels of the device.
    */
-  QDMI_DEVICE_PULSE_SUPPORT_LEVEL_SITEANDCHANNEL = 3,
+  QDMI_DEVICE_PULSE_SUPPORT_LEVEL_SITEANDCHANNEL = 3
```
```suggestion
  QDMI_DEVICE_PULSE_SUPPORT_LEVEL_SITEANDCHANNEL = 3,
```
seems unintentional

include/qdmi/constants.h
```diff
@@ -512,6 +512,15 @@ enum QDMI_OPERATION_PROPERTY_T {
   QDMI_OPERATION_PROPERTY_DURATION = 3,
   /// `double` The fidelity of an operation.
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
+  /**
+   * @brief @ref QDMI_Operation_Scope The scope of the operation.
+   * @details The scope of an operation indicates whether the operation
+   * executes on individual sites or on all sites this operation can be applied
+   * to.
+   * @note This property is mainly required for neutral atom devices that offer
+   * apart from local operations also global operations.
```
```suggestion
   * @note This property is mainly required for neutral atom devices that offer
   * global operations in addition to local operations.
```
The original wording feels awkward.

test/test_qdmi.cpp
```diff
@@ -168,6 +168,12 @@ TEST_P(QDMIImplementationTest, QueryGatePropertiesForEachGate) {
                   QDMI_OPERATION_PROPERTY_MAX, 0, nullptr, nullptr),
               QDMI_ERROR_INVALIDARGUMENT);
 
+    // The example devices do not support neutral atom-specific properties
+    EXPECT_EQ(QDMI_device_query_operation_property(
+                  device, op, 0, nullptr, 0, nullptr,
+                  QDMI_OPERATION_PROPERTY_SCOPE, 0, nullptr, nullptr),
+              QDMI_ERROR_NOTSUPPORTED);
```
burgholzer: Instead of that, I'd rather have the example device report all operations as local.
It's easy to add and it does not hurt to provide complete information.

burgholzer (changes requested):
Many thanks also here. Again, a couple of minor (hopefully) comments.

include/qdmi/constants.h
ystade: Yeah, I had a similar line of thought. When the same operation can individually address multiple zones (haven't seen that so far), this operation would have to be split into separate distinct operations each acting on one zone.

One addition: `global` in this case here means: global on all sites the operation can act on, not global to the device. And yes, you are right, those sites are returned as a 1D list.

ystade: And yes, for now a bollean flag would be sufficient: Should it be called `...GLOBAL` or `...LOCAL` then?

ystade: Ok the last commits contain the proposal with a boolean flag now. What do you think?

burgholzer: > Yeah, I had a similar line of thought. When the same operation can individually address multiple zones (haven't seen that so far), this operation would have to be split into separate distinct operations each acting on one zone.

> One addition: global in this case here means: global on all sites the operation can act on, not global to the device. And yes, you are right, those sites are returned as a 1D list.

But then I am not yet quite sure how this would work at the moment.
Probably I am really not seeing something here, but let me note the following and you tell me where I am going wrong:
In the current implementation, there is actually no way to ask an individual operation which sites it can act on.
What we currently [do in the tests](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/b3c980ceaa435121f72125b3eafba8c21e1c6ed7/test/test_qdmi.cpp#L126-L163) is that
- we iterate over all sites for single qubit gates and query the respective properties we want (which would return NOTSUPPORTED if the operation is not supported on that site)
- we iterate over all pairs of sites in the coupling map and perform a similar query.

For neutral atom platforms, the coupling map does not really make sense.
For single-qubit gates, I would query properties for an individual site. Assuming the operation on that site reports that it is global. How would I know which other sites this affects?
Similarly for two-qubit gates, where I don't even know over what to iterate.
Is the problem coming across?

include/qdmi/constants.h
```diff
@@ -513,6 +513,13 @@ enum QDMI_OPERATION_PROPERTY_T {
   QDMI_OPERATION_PROPERTY_DURATION = 3,
   /// `double` The fidelity of an operation.
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
+  /**
+   * @brief `bool` An operation is global if it acts on all sites
+   * simultaneously. the operation is applicable to.
```
```suggestion
   * @brief `bool` An operation is global if it acts on all sites
   * the operation is applicable to simultaneously.
```
Something was not quite right here.

include/qdmi/constants.h
burgholzer: > Ok the last commits contain the proposal with a boolean flag now. What do you think?

I like that 👍🏼

ystade: I see your problem. The fact is that the actual operation is still a single-qubit operation, hence, it makes sense to ask for individual sides whether an operation can be applied to that site or not. When the operation reports that it is global, then you know that it also acts on all other sites where the operation did not report UNSUPPORTED.

Similarly for two-qubit gates: You iterate over all (!) pairs of sites. If an operation is global then it acts on all pairs of sites that did not report UNSUPPORTED simultaneously when executed.

Doe sthis make sense?

burgholzer: Based on offline discussions, this will probably require a couple more thoughts. Resolving this thread for now. Let's start a new one based on the first proposed solution.

ystade: Ok, you'll find the answer in the main thread soon and I'll resolve that now.

ystade:
After a longer discussion, this PR was significantly re-structured. I now updated the PR description to fit the new design and also put the motivation into the `rationale.md` file.

@burgholzer Suprisingly, even though it feels like a major change, the actual changes made to the code are very minor. Feel free to look at the code again and give your feedback.

include/qdmi/constants.h
```diff
@@ -468,6 +468,62 @@ enum QDMI_SITE_PROPERTY_T {
...
+   * @note This property is a length value and must be interpreted in the way
+   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
+   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
+   * @note This property is mainly required for neutral atom devices to
+   * report the extent of zones, see @ref QDMI_SITE_PROPERTY_ISZONE.
+   * @note If the site is not a zone, this property must return @ref
+   * QDMI_ERROR_NOTSUPPORTED.
+   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
+   */
+  QDMI_SITE_PROPERTY_ZEXTENT = 10,
```
```suggestion
  /**
   * @brief `uint64_t` The extent of a zone along the X-axis.
   * @note This property is a length value and must be interpreted in the way
   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
   * @note This property is mainly required for neutral atom devices to
   * report the extent of zones, see @ref QDMI_SITE_PROPERTY_ISZONE.
   * @note If the site is not a zone, this property must return @ref
   * QDMI_ERROR_NOTSUPPORTED.
   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
   */
  QDMI_SITE_PROPERTY_XEXTENT = 8,
  /**
   * @brief `uint64_t` The extent of a zone along the Y-axis.
   * @note This property is a length value and must be interpreted in the way
   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
   * @note This property is mainly required for neutral atom devices to
   * report the extent of zones, see @ref QDMI_SITE_PROPERTY_ISZONE.
   * @note If the site is not a zone, this property must return @ref
   * QDMI_ERROR_NOTSUPPORTED.
   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
   */
  QDMI_SITE_PROPERTY_YEXTENT = 9,
  /**
   * @brief `uint64_t` The extent of a zone along the Z-axis.
   * @note This property is a length value and must be interpreted in the way
   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
   * @note This property is mainly required for neutral atom devices to
   * report the extent of zones, see @ref QDMI_SITE_PROPERTY_ISZONE.
   * @note If the site is not a zone, this property must return @ref
   * QDMI_ERROR_NOTSUPPORTED.
   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
   */
  QDMI_SITE_PROPERTY_ZEXTENT = 10,
```

docs/rationale.md
```diff
@@ -227,3 +227,22 @@ the same enumeration, each with a different prefix. It would also not be possibl
...
+realized that this was not general enough. These devices offer _global_ operations that execute on
+multiple atoms within a given zone simultaneously. In particular, these global operations cannot be
+executed on individual atoms but on all atoms within a zone at once. Consequently, we had to equip
+QDMI such that it is capable of reporting operation properties with respect to zones and not only
+individual qubit positions. To this end, we introduced the concept of _zone sites_ that represent a
+spatial zone. These have the same type as regular sites, i.e., @ref QDMI_Site, but are used to
+represent a zone instead of a single qubit position. Hence, they can be used in the function @ref
+QDMI_device_query_operation_property as parameter to query data like fidelity or duration of global
+operations specific to a zone. This allows QDMI to represent the capabilities of neutral atom-based
+devices while keeping the interface consistent with the existing site concept.
```
burgholzer: Should this be mentioning something about the fact that the site query function is expected to report all sites, i.e., individual sites as well as zones?
Right now, it is up to the levels above QDMI to query the sites and split them between zones and non-zones, right?

burgholzer (changes requested):
This looks pretty good already. The fact that it's just a couple of changes is a good sign in my opinion.

docs/rationale.md
ystade: Yes, this is so to say the minimal invasive option where the levels above QDMI have to filter out the zone sites or the regular sites from all sites. the device property yields all sites.

ystade: In the last commit I made the docstring clearer.

burgholzer (approved):
This LGTM now 🙌🏼 Feel free to merge if you are satisfied as well.

PR #200, opened 2025-08-05
✨ Add Module and Submodule Index

### LINKED ISSUE #189: ✨ Representing Neutral Atom Arrays

ystade:
## Description

This PR realises the changes proposed in #189.

Fixes #189

## Checklist:

### DIFF
--- a/examples/device/device.cpp
+++ b/examples/device/device.cpp
@@ -796,6 +796,8 @@ int CXX_QDMI_device_session_query_site_property(CXX_QDMI_Device_Session session,
   }
   ADD_SINGLE_VALUE_PROPERTY(QDMI_SITE_PROPERTY_INDEX, uint64_t, site->id, prop,
                             size, value, size_ret)
+  ADD_SINGLE_VALUE_PROPERTY(QDMI_SITE_PROPERTY_MODULEINDEX, uint64_t, 0, prop,
+                            size, value, size_ret)
   ADD_SINGLE_VALUE_PROPERTY(QDMI_SITE_PROPERTY_T1, double, 1000.0, prop, size,
                             value, size_ret)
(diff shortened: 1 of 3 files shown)


burgholzer (approved):
This looks reasonable to me! 🙌

PR #203, opened 2025-08-07
✨ Reporting Interaction and Blocking Radius

### LINKED ISSUE #202: ✨ Reporting Interaction and Blocking Radius

As mentioned in other issues (#185, #189), neutral atom-based devices do not hold the atoms representing qubits in fixed positions. Hence, it is not feasible to define a coupling map for those devices. To define which atoms can interact for, e.g., CZ-gates, those devices define an interaction radius and a blocking radius, where the blocking radius is greater than the interaction radius. All atoms within the interaction radius will interact. Atoms that are not supposed to each other should be kept further away than the blocking radius to prevent cross-talk.

So far, QDMI does not offer any possibility to report those two figures.

Since these figures mentioned above are not specific to certain sites, they can be reported globally for the entire device. However, they may depend on the actual operation. In principle, there could be different operations featuring different interactions and blocking radii. Additionally, the interaction and blocking radius is only relevant for entangling multi-qubit gates, like CZ-gates.

Hence, I propose to add both figures as operation properties.

ystade:
## Description

This PR realizes #202.

Fixes #202

## Checklist:

### DIFF
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -588,7 +588,7 @@ enum QDMI_OPERATION_PROPERTY_T {
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
   /**
    * @brief `uint64_t` The interaction radius of the operation.
-   * @details The interaction radius is the maximum distance between two 
+   * @details The interaction radius is the maximum distance between two
    * qubits that can be involved in the operation. It only applies to
    * multi-qubit gates.
    * @note This property is mainly required for neutral atom devices where
@@ -607,7 +607,7 @@ enum QDMI_OPERATION_PROPERTY_T {
    * qubits that should not be involved in the operation to avoid crosstalk.
    * It only applies to multi-qubit gates.
    * @note This property is mainly required for neutral atom devices where
-   * atoms representing qubits can be at arbitrary locations. To avoid 
+   * atoms representing qubits can be at arbitrary locations. To avoid
    * crosstalk, the blocking radius of the operation must be respected when
    * scheduling operations.
    * @note This property is a length value and must be interpreted in the way
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -587,6 +587,36 @@ enum QDMI_OPERATION_PROPERTY_T {
   /// `double` The fidelity of an operation.
   QDMI_OPERATION_PROPERTY_FIDELITY = 4,
   /**
+   * @brief `uint64_t` The interaction radius of the operation.
+   * @details The interaction radius only applies to multi-qubit gates. The
+   * interaction radius is the maximum distance between two qubits that can be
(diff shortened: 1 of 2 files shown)


include/qdmi/constants.h
```diff
@@ -586,6 +586,36 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+   * @note This property is a length value and must be interpreted in the way
+   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
+   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
+   * @note This property is mainly required for neutral atom devices where
+   * atoms representing qubits can be at arbitrary locations. Hence, it is
+   * infeasible to define a coupling map. To avoid crosstalk, the blocking
+   * radius of the operation must be respected.
+   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
+   */
+  QDMI_OPERATION_PROPERTY_BLOCKINGRADIUS = 6,
```
```suggestion
  /**
   * @brief `uint64_t` The interaction radius of the operation.
   * @details The interaction radius is the maximum distance between two 
   * qubits that can be involved in the operation. It only applies to
   * multi-qubit gates.
   * @note This property is mainly required for neutral atom devices where
   * atoms representing qubits can be at arbitrary locations. Hence, it is
   * infeasible to define a coupling map. Instead, the coupling of atoms is
   * defined by the interaction radius of the operation.
   * @note This property is a length value and must be interpreted in the way
   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
   */
  QDMI_OPERATION_PROPERTY_INTERACTIONRADIUS = 5,
  /**
   * @brief `uint64_t` The blocking radius of the operation.
   * @details The blocking radius is the minimum distance between two
   * qubits that should not be involved in the operation to avoid crosstalk.
   * It only applies to multi-qubit gates.
   * @note This property is mainly required for neutral atom devices where
   * atoms representing qubits can be at arbitrary locations. To avoid 
   * crosstalk, the blocking radius of the operation must be respected when
   * scheduling operations.
   * @note This property is a length value and must be interpreted in the way
   * specified by the @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT and @ref
   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR properties.
   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
   */
  QDMI_OPERATION_PROPERTY_BLOCKINGRADIUS = 6,
```
Just a suggestion for a slight refactoring.

burgholzer (approved):
LGTM. Just a small suggestion on the docstring wording.

PR #207, opened 2025-08-10
✨ Efficiently Query Connectivity/Applicability of Operations

### LINKED ISSUE #206: ✨ Efficiently Query Connectivity/Applicability of Operations

To find which qubits an operation can be applied to the current workflow looks as follows:
1. query all sites via the corresponding device property
1. query all operations via the corresponding device property
1. find the operation of interest, i.e., by filtering by name
1. call the function `QDMI_session_query_operation_property` with the operation of interest and any combination of a site; if the result is `QDMI_ERROR_NOTSUPPORTED` the operation cannot be applied to this site, if the result is `QDMI_SUCCESS` the operation can be applied to the site.

For step 4., one also has to specify a property that one queries even though one is not interested in the actual value of the property. For this, one needs a property that must be supported, such that `QDMI_ERROR_NOTSUPPORTED` really indicates that the given site is the cause for the error value. So, one would likely choose `QDMI_OPERATION_PROPERTY_NAME` for this purpose.

However, this does not feel really ergonomic. Additionally, the procedure above was assuming that the operation of interest is a single-qubit operation; hence, we just have to provide one site per call. Already then, we have to call the `QDMI_session_query_operation_property` n times where n is the number of sites.

This gets even worse when we have a two-qubit or even higher multi-qubit gate. Then we have to query the operation with all combinations, which are n^2 or n^k many, respectively.

To circumvent the high number of calls to the query function, a new `QDMI_OPERATION_PROPERTY_SITES` could return a list of all supported sites the operation can be applied to. In particular, for an operation where `NUMQUBITS` is one, this will just be a list of all supported individual sites. For operation where `NUMQUBITS` is 2, this will be a list of pairs meaning that the operation is supported on all pairs of the list. This can be generalized to 3-qubit, 4-qubit, ... gates.

The only two concerns I have here are the following:
1. This property is redundant because the respective information can already be retrieved as described above in the first section.
2. The list returned, especially, when containing pairs of triples may be very long and consume a lot of memory.

ystade:
## Description

Implements the necessary feature from #206.

Fixes #206

## Checklist:

### DIFF
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -645,14 +645,17 @@ enum QDMI_OPERATION_PROPERTY_T {
    */
   QDMI_OPERATION_PROPERTY_IDLINGFIDELITY = 7,
   /**
-   * @brief `bool` Whether the operation is a global operation.
-   * @details A global operation is an operation that can be applied to all
-   * qubits in a zone simultaneously.
-   * @note The default value is `false`. If an operation reports @ref
-   * QDMI_ERROR_NOTSUPPORTED for this property, it is assumed to be a local (not
-   * global) operation.
+   * @brief `bool` Whether the operation is a zoned (global) operation.
+   * @details A zoned (or global) operation is an operation that can be applied
+   * simultaneously to all qubits within a specific zone. If this property is
+   * `true`, the operation is considered zoned. If it is `false` or returns @ref
+   * QDMI_ERROR_NOTSUPPORTED, the operation is considered local. The
+   * applicability of a zoned operation to specific zones is detailed in @ref
+   * QDMI_OPERATION_PROPERTY_SITES.
    * @note This property is primarily relevant for neutral atom devices, where a
    * laser can illuminate an entire array of atoms representing qubits.
+   * @see QDMI_SITE_PROPERTY_ISZONE
+   * @see QDMI_OPERATION_PROPERTY_SITES
    */
   QDMI_OPERATION_PROPERTY_ISZONED = 8,
   /**
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -617,6 +617,21 @@ enum QDMI_OPERATION_PROPERTY_T {
    */
   QDMI_OPERATION_PROPERTY_BLOCKINGRADIUS = 6,
   /**
+   * @brief `QDMI_Site*` (@ref QDMI_Site list) The sites the operation is
+   * applicable to.
+   * @details The operations can be applied to every site of the list of
+   * returned sites. In case of multi-qubit operations the returned list will be
+   * a list of tuples. More precisely, for a k-qubit operation, where k is the
+   * number returned @ref QDMI_OPERATION_PROPERTY_QUBITSNUM, the list consists
+   * of k-tuples. The semantics is, that the operations can be applied to every
+   * tuple returned in the list. In case of global operations, the property will
+   * return a list of zone sites.
+   * @par
+   * The returned @ref QDMI_Site handles must be a subset of the sites returned
+   * by @ref QDMI_DEVICE_PROPERTIES_SITES.
+   */
+  QDMI_OPERATION_PROPERTY_SITES = 7,
+  /**
    * @brief The maximum value of the enum.
    * @details It can be used by devices for bounds checking and validation of
    * function parameters.
@@ -624,7 +639,7 @@ enum QDMI_OPERATION_PROPERTY_T {
    * @attention This value must remain the last regular member of the enum
    * besides the custom members and must be updated when new members are added.
    */
-  QDMI_OPERATION_PROPERTY_MAX = 7,
+  QDMI_OPERATION_PROPERTY_MAX = 8,
   /**
    * @brief This enum value is reserved for a custom property.
    * @details The device defines the meaning and the type of this property.
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -628,7 +628,7 @@ enum QDMI_OPERATION_PROPERTY_T {
    * return a list of zone sites.
    * @par
    * The returned @ref QDMI_Site handles must be a subset of the sites returned
-   * by @ref QDMI_DEVICE_PROPERTIES_SITES.
+   * by @ref QDMI_DEVICE_PROPERTY_SITES.
    */
   QDMI_OPERATION_PROPERTY_SITES = 7,
   /**
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -626,16 +626,16 @@ enum QDMI_OPERATION_PROPERTY_T {
    * @note This property is primarily relevant for neutral atom devices, where a
    * laser can illuminate an entire array of atoms representing qubits.
    */
-  QDMI_OPERATION_PROPERTY_ISGLOBAL = 7,
+  QDMI_OPERATION_PROPERTY_ISZONED = 7,
   /**
    * @brief `QDMI_Site*` (list) The sites to which the operation is applicable.
    * @details
-   * - For local operations (see @ref QDMI_OPERATION_PROPERTY_GLOBAL), this
+   * - For local operations (see @ref QDMI_OPERATION_PROPERTY_ISZONED), this
    * property returns a list of tuples. Each tuple contains sites from the list
    * provided by @ref QDMI_DEVICE_PROPERTY_SITES and represents a valid
    * combination for the operation. The number of sites in each tuple matches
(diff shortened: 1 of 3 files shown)


ystade:
@burgholzer I have now implemented the new property returning a list, see also docstring. Nevertheless, I have one question: For the set I am using in the tests and where I store pairs of sites, I have to define their hash function. Is there a better solution than the current? Since it is implemented in the global namespace, will that conflict with symbols from other libraries that use QDMI?

include/qdmi/constants.h
```diff
@@ -616,6 +616,21 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+   * a list of tuples. More precisely, for a k-qubit operation, where k is the
+   * number returned @ref QDMI_OPERATION_PROPERTY_QUBITSNUM, the list consists
+   * of k-tuples. The semantics is, that the operations can be applied to every
+   * tuple returned in the list. In case of global operations, the property will
+   * return a list of zone sites.
+   * @par
+   * The returned @ref QDMI_Site handles must be a subset of the sites returned
+   * by @ref QDMI_DEVICE_PROPERTY_SITES.
+   */
+  QDMI_OPERATION_PROPERTY_SITES = 7,
```
burgholzer: I don't think this is quite clear yet.
How about something along the lines of 

```
/**
 * @brief `QDMI_Site*` (@ref QDMI_Site list) The sites to which the operation is applicable.
 *
 * @details
 * When queried, this property returns a list of tuples. Each tuple contains sites from the list
 * provided by @ref QDMI_DEVICE_PROPERTY_SITES and represents a valid combination on which the operation
 * can be applied. The number of sites in each tuple equals the value returned by
 * @ref QDMI_OPERATION_PROPERTY_QUBITSNUM.
 *
 * For multi-qubit operations, every tuple produced corresponds to the physical sites used for the operation.
 * For global operations, the returned list contains zone sites, i.e., zones where the operation can be applied.
 */
```

What I do not yet like very much about this: A global multi-qubit operation will currently return `QDMI_OPERATION_PROPERTY_QUBITSNUM > 1`, which somewhat conflicts with the return value here being a list of zone sites and the general docstring, which is why we have to explicitly mention the exception for global operations.
At the same time, if I am not missing anything, we currently do not have an easy way to ask an operation if it is a "zoned"/global operation. So it is not particularly easy to programmatically make that semantic exception at the moment.
Looking forward to your input on this.

test/test_qdmi.cpp
```diff
@@ -25,16 +25,27 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
 #include <vector>
 
+/// Hash function for a pair
+template <class T, class U> struct std::hash<std::pair<T, U>> {
+  auto operator()(const std::pair<T, U> &p) const noexcept -> std::size_t {
+    const std::size_t h1 = std::hash<T>{}(p.first);
+    const std::size_t h2 = std::hash<U>{}(p.second);
+    return h1 ^ h2;
+  }
+};
```
burgholzer: To avoid the manipulation of the global namespace here, you can make use of the third template argument of the `unordered_map` class (https://en.cppreference.com/w/cpp/container/unordered_map/unordered_map.html), which is the function being used for hashing.

test/test_qdmi.cpp
```diff
@@ -144,7 +171,27 @@ TEST_P(QDMIImplementationTest, QueryGatePropertiesForEachGate) {
...
+      ASSERT_EQ(QDMI_device_query_operation_property(
+                    device, op, 0, nullptr, 0, nullptr,
+                    QDMI_OPERATION_PROPERTY_SITES, size_of_supported_site_pairs,
+                    supported_site_pairs.data(), nullptr),
+                QDMI_SUCCESS);
+      for (const auto &pair : supported_site_pairs) {
+        EXPECT_NE(set_of_all_site_pairs.find(pair), set_of_all_site_pairs.end())
+            << "Supported site pairs must be a subset of edges in the coupling "
+               "map.";
+        const auto &[control, target] = pair;
```
burgholzer: This only applies to devices that support the definition of a coupling map.
Should this test be conditional on whether the coupling map property is available? If it is not available, then only the subset condition from the single-qubit case above is checked?

burgholzer (changes requested):
This might need another round, but then it should be ready.

include/qdmi/constants.h
ystade: I had a similar thought already and asked myself whether we should revisit #199 and add a `GLOBAL`property to the operation. Should I incoporate this here?

test/test_qdmi.cpp
ystade: Honestly, I completely agree. However, the current test code is partially quite specific to the example device anyways. So, I personally would be fine if this is not general either and remains this way.

include/qdmi/constants.h
burgholzer: I think it makes sense to add such a property. Might as well do it here since this is where the need for it became apparent.
The only question for me is whether this should be called `ISGLOBAL` or `ISZONED` or, the inverse, `ISLOCAL`.
We currently refer to the sites that are not local as `Zones`. So "global" does not seem quite like a 100% match.
The important thing is that this property should be optional and the default should imply that every operation is local.

Based on whatever comes out of the above, the docstring here can then be updated.

test/test_qdmi.cpp
burgholzer: Fine with me as well.
At some point, we will probably want to create separate sets of tests. Some that must work for all conceivable QDMI devices and other sets of tests that each only work for particular devices but really test the devices themselves.
Not in this PR though.

include/qdmi/constants.h
ystade: True, I prefer is `ISZONED` because then the default can still be `false`, which feels a bit more natural than a default that is `true`.

burgholzer: `ISZONED` works for me 👍🏼

test/test_qdmi.cpp
```diff
@@ -125,7 +138,23 @@ TEST_P(QDMIImplementationTest, QueryGatePropertiesForEachGate) {
...
     double fidelity = 0;
     if (gate_num_qubits == 1) {
-      for (const auto &site : sites) {
+      const std::unordered_set set_of_all_sites(sites.cbegin(), sites.cend());
+      size_t size_of_supported_sites = 0;
+      ASSERT_EQ(QDMI_device_query_operation_property(
+                    device, op, 0, nullptr, 0, nullptr,
+                    QDMI_OPERATION_PROPERTY_SITES, 0, nullptr,
+                    &size_of_supported_sites),
+                QDMI_SUCCESS);
```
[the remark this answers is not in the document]
ystade: While true in general, I find this test too fine-grained. So, I'll resolve without further ado.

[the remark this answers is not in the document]
ystade: Or, why not, it does not hurt...

include/qdmi/constants.h
```diff
@@ -644,6 +644,30 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+  /**
+   * @brief `bool` Whether the operation is a global operation.
+   * @details A global operation is an operation that can be applied to all
+   * qubits in a zone simultaneously.
+   * @note The default value is `false`. If an operation reports @ref
+   * QDMI_ERROR_NOTSUPPORTED for this property, it is assumed to be a local (not
+   * global) operation.
+   * @note This property is primarily relevant for neutral atom devices, where a
+   * laser can illuminate an entire array of atoms representing qubits.
+   */
```
burgholzer: I don't quite like that this is now called `ISZONED`, but the brief description only uses "global". How about the following:
```
/**
 * @brief `bool` Whether the operation is a zoned (global) operation.
 * @details A zoned operation is a global operation that can be applied simultaneously to all qubits within a specific zone.
 * If this property is `true`, the operation is considered zoned. If it is `false` or returns @ref QDMI_ERROR_NOTSUPPORTED, the operation is considered local.
 * The applicability of a zoned operation to specific zones is detailed in @ref QDMI_OPERATION_PROPERTY_SITES.
 * @note This property is primarily relevant for neutral atom devices, where a laser can illuminate an entire array of atoms representing qubits.
 * @see QDMI_SITE_PROPERTY_ISZONE
 * @see QDMI_OPERATION_PROPERTY_SITES
 */
QDMI_OPERATION_PROPERTY_ISZONED = 8,
```

burgholzer (approved):
This looks good to me now besides one suggestion for the `ISZONED` property. Feel free to adapt that as you see fit and merge this afterwards.

include/qdmi/constants.h
ystade: Good catch!

PR #208, opened 2025-08-10
 ✨ Report the Fidelity of Idling Qubits under a Rydberg Beam

### LINKED ISSUE #205: ✨ Report the Fidelity of Idling Qubits under a Rydberg Beam

Neutral atom devices have the special feature of global laser beams that illuminate entire arrays of atoms, inducing an operation on them. For multi-qubit gates, such as the CZ-gate, the illuminated atoms are excited to a high-lying energy state. Atoms within the interaction radius (#202) then interact and execute a CZ-gate. However, even individual atoms are excited to that high-lying energy state and then just decay again without interacting with any atom, performing any gate. However, this decay induces a lower fidelity than just idling without being affected by the laser beam. Hence, there is a certain fidelity next to the fidelity of the operation applicable to idling qubits within the zone addressed by this laser beam. This fidelity is usually called _idling fidelity_. Right now, there is no way to report the idling fidelity through QDMI since it is an additional number to the operation's fidelity.

One possibility I see is to add another operation property `QDMI_OPERATION_PROPERTY_IDLINGFIDELITY` that reports the number described above. Operations for which this is not relevant can report `NOT_SUPPORTED`.

Another one could be to introduce a second operation whose fidelity is exactly the idling fidelity of the first operation. However, I do not see right now how to relate those operations then meaningfully in QDMI.

ystade:
## Description

Implements the feature described in #205.

Fixes #205

## Checklist:

### DIFF
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -619,12 +619,13 @@ enum QDMI_OPERATION_PROPERTY_T {
   /**
    * @brief `double` The idling fidelity of an operation.
    * @details Additional to @ref QDMI_OPERATION_PROPERTY_FIDELITY, this property
-   * denotes the fidelity on qubits that are affected by this operation but do
-   * not execute this operation. E.g., when applying a global multi-qubit
-   * operation, also isolated qubits that do not have another atom in their
-   * interaction radius to interact with, are affected. Even though they undergo
-   * an identity operation, this may be faulty and has higher infidelity than
-   * just idling not affected by the operation.
+   * denotes the fidelity on qubits that are affected by a global  operation but
+   * do not execute this operation because they do not have an interaction
+   * partner. When applying a global multi-qubit operation, also isolated qubits
+   * that do not have another atom in their interaction radius to interact with,
+   * are affected. Even though they undergo an identity operation, this may be
+   * faulty and has higher infidelity than just idling not affected by the
+   * operation.
    * @note This property is mainly required for neutral atom devices where
    * operations are performed by shining lasers on qubits. These can also
    * illuminate an entire array of atoms. This affects all atoms in the array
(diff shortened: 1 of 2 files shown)


include/qdmi/constants.h
```diff
@@ -616,6 +616,22 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+   * that do not have another atom in their interaction radius to interact with,
+   * are affected. Even though they undergo an identity operation, this may be
+   * faulty and has higher infidelity than just idling not affected by the
+   * operation.
+   * @note This property is mainly required for neutral atom devices where
+   * operations are performed by shining lasers on qubits. These can also
+   * illuminate an entire array of atoms. This affects all atoms in the array
+   * even those that do not interact and hence do not execute the operation.
+   */
+  QDMI_OPERATION_PROPERTY_IDLINGFIDELITY = 7,
```
```suggestion
  /**
   * @brief `double` Fidelity of qubits idling during a global operation.
   * @details This property measures the fidelity of qubits that are within the affected area of a global multi-qubit operation but do not actively participate (i.e., they lack an interaction partner within their radius). Even though these qubits undergo an identity operation, errors may still occur, resulting in lower fidelity compared to qubits that are simply idling and not exposed to the operation.
   * @note This is especially relevant for neutral atom devices, where global operations (e.g., laser pulses) can impact all atoms in the array, including those not interacting.
   */  
  QDMI_OPERATION_PROPERTY_IDLINGFIDELITY = 7,
```
Quickly ran this through Copilot and I quite like the description (which will be reformated by clang-format). Feel free to tweak it further if there is something you do not like.

Maybe also worth to tag Copilot for a review after incorporating the suggestion. Just to see if it finds anything else to improve.

burgholzer (approved):
This looks good already. I left a suggestion for the docstring that might be worth considering.

PR #210, opened 2025-08-10
✨ Introducing Duration Unit and Reporting Shuttling Speed

### LINKED ISSUE #191: ✨ Reporting shuttling speed

Neutral atom devices have the ability to dynamically rearrange atoms during computation. Naturally, the duration of the rearrangement, also called shuttling, depends on the distance the atoms travel. In the simplest model, the atoms are moved with a constant speed. In reality, however, they are accelerated and slowed down with a constant jerk (derivation of the acceleration). Additionally, there might be a maximum speed.

I see two possibilities, that might be combined:
1. Reporting the (average) speed atoms can be shuttled. This can be used to calculate the duration of a rearrangement.
2. Reporting the maximum jerk and maximum speed the device allows for shuttling atoms.
For all properties, correspinding new entries would have to be added to the `QDMI_DEVICE_PROPERTIES` enum.

ystade:
## Description

This PR adds properties to report the duration unit used by the device, similarly to the length unit introduced in #198.

On top, it adds the feature to report the shuttling speed requested in #191.

Fixes #191

## Checklist:

### DIFF
--- a/examples/fomac/example_fomac.cpp
+++ b/examples/fomac/example_fomac.cpp
@@ -166,20 +166,45 @@ auto FoMaC::get_site_id(QDMI_Site site) const -> uint64_t {
   return site_id;
 }
 
+auto FoMaC::get_us_scale_factor() const -> double {
+  size_t size = 0;
+  int ret = QDMI_device_query_device_property(
+      device, QDMI_DEVICE_PROPERTY_DURATIONUNIT, 0, nullptr, &size);
+  throw_if_error(ret, "Failed to retrieve the device's duration unit size.");
+  std::string unit(size - 1, '\0');
+  ret = QDMI_device_query_device_property(
+      device, QDMI_DEVICE_PROPERTY_DURATIONUNIT, unit.size() + 1, unit.data(),
+      nullptr);
+  throw_if_error(ret, "Failed to retrieve the device's duration unit.");
+  double scale_factor = 0.0;
+  ret = QDMI_device_query_device_property(
+      device, QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR, sizeof(double),
+      &scale_factor, nullptr);
+  throw_if_error(ret, "Failed to query the duration scale factor");
+  if (unit == "ms") {
+    scale_factor /= 1000;
+  } else if (unit == "ns") {
+    scale_factor *= 1000;
+  } else if (unit != "us") {
+    throw std::runtime_error("Unrecognized unit: " + unit);
+  }
+  return scale_factor;
+}
+
 auto FoMaC::get_site_t1(QDMI_Site site) const -> double {
-  double t1 = 0;
+  uint64_t t1 = 0;
   const int ret = QDMI_device_query_site_property(
       device, site, QDMI_SITE_PROPERTY_T1, sizeof(double), &t1, nullptr);
   throw_if_error(ret, "Failed to query the T1 time");
-  return t1;
+  return t1 * get_us_scale_factor();
 }
 
 auto FoMaC::get_site_t2(QDMI_Site site) const -> double {
-  double t2 = 0;
+  uint64_t t2 = 0;
   const int ret = QDMI_device_query_site_property(
       device, site, QDMI_SITE_PROPERTY_T2, sizeof(double), &t2, nullptr);
   throw_if_error(ret, "Failed to query the T2 time");
-  return t2;
+  return t2 * get_us_scale_factor();
 }
 
 auto FoMaC::get_operands_num(const QDMI_Operation &op) const -> size_t {
--- a/examples/fomac/example_fomac.cpp
+++ b/examples/fomac/example_fomac.cpp
@@ -194,7 +194,7 @@ auto FoMaC::get_us_scale_factor() const -> double {
 auto FoMaC::get_site_t1(QDMI_Site site) const -> double {
   uint64_t t1 = 0;
   const int ret = QDMI_device_query_site_property(
-      device, site, QDMI_SITE_PROPERTY_T1, sizeof(double), &t1, nullptr);
+      device, site, QDMI_SITE_PROPERTY_T1, sizeof(uint64_t), &t1, nullptr);
   throw_if_error(ret, "Failed to query the T1 time");
   return static_cast<double>(t1) * get_us_scale_factor();
 }
@@ -202,7 +202,7 @@ auto FoMaC::get_site_t1(QDMI_Site site) const -> double {
 auto FoMaC::get_site_t2(QDMI_Site site) const -> double {
   uint64_t t2 = 0;
   const int ret = QDMI_device_query_site_property(
-      device, site, QDMI_SITE_PROPERTY_T2, sizeof(double), &t2, nullptr);
+      device, site, QDMI_SITE_PROPERTY_T2, sizeof(uint64_t), &t2, nullptr);
   throw_if_error(ret, "Failed to query the T2 time");
   return static_cast<double>(t2) * get_us_scale_factor();
 }
--- a/examples/fomac/example_fomac.cpp
+++ b/examples/fomac/example_fomac.cpp
@@ -196,7 +196,7 @@ auto FoMaC::get_site_t1(QDMI_Site site) const -> double {
   const int ret = QDMI_device_query_site_property(
       device, site, QDMI_SITE_PROPERTY_T1, sizeof(double), &t1, nullptr);
   throw_if_error(ret, "Failed to query the T1 time");
-  return t1 * get_us_scale_factor();
+  return static_cast<double>(t1) * get_us_scale_factor();
 }
 
 auto FoMaC::get_site_t2(QDMI_Site site) const -> double {
@@ -204,7 +204,7 @@ auto FoMaC::get_site_t2(QDMI_Site site) const -> double {
   const int ret = QDMI_device_query_site_property(
       device, site, QDMI_SITE_PROPERTY_T2, sizeof(double), &t2, nullptr);
   throw_if_error(ret, "Failed to query the T2 time");
-  return t2 * get_us_scale_factor();
+  return static_cast<double>(t2) * get_us_scale_factor();
 }
 
 auto FoMaC::get_operands_num(const QDMI_Operation &op) const -> size_t {
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -356,7 +356,7 @@ enum QDMI_DEVICE_PROPERTY_T {
   QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR = 11,
   /**
    * @brief `char*` (string) The duration unit used by the device.
-   * @details This property must be a known SI unit, e.g., "s", "ms" or "ns".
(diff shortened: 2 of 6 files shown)


ystade:
@burgholzer While implementing the support for the shuttling speed (mean shuttling speed for now) as requested in #191, I realized that the unit of the shuttling speed is a combination of the length and duration units. The current state of this PR, hence, adds similarly to the length unit, the properties for a duration unit. As mentioned in #185 ([here](https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/185#issuecomment-3032103836)), the introduction of the duration unit is a breaking change, as all properties related to duration values must be adapted where there semantics used a unit fixed by QDMI previously. Consequently, I tagged this PR as `Minor`. Let me know, whether we want to add those features now for the shuttling speed or whether we want to keep it simple for now and also just fix the unit for the shuttling speed to um/us in the QDMI docstring.

include/qdmi/constants.h
```diff
@@ -354,6 +354,24 @@ enum QDMI_DEVICE_PROPERTY_T {
...
+  QDMI_DEVICE_PROPERTY_DURATIONUNIT = 12,
+  /**
+   * @brief `double` A factor applied to all duration values.
+   * @details This value must be multiplied with all duration values returned by
+   * the device before it can be interpreted in the unit returned by  @ref
+   * QDMI_DEVICE_PROPERTY_DURATIONUNIT.
+   * @note If this property is @ref QDMI_ERROR_NOTSUPPORTED, a default value of
+   * `1.0` is assumed.
+   */
+  QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR = 13,
```
burgholzer: I am really sorry, but I keep getting confused with these docstrings about who's responsibility it is to do the scaling computation purely based on the docstrings and without looking at the example device. This was already the case for the length scale and unit.

I had a conversation with Gemini on this, which led to the following suggestions for improvements of the docstrings here

```c++
/**
 * @brief `char*` (string) The duration unit reported by the device.
 * @details The device implementation must report a known SI unit (e.g., "ms", "us", or "ns") for this property.
 * A client querying a duration value must first scale it using @ref QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR.
 * The resulting value is then interpreted in the unit specified by this property.
 * @note If the device reports any duration values, this property must be set.
 */
QDMI_DEVICE_PROPERTY_DURATIONUNIT = 12,
```

```c++
/**
 * @brief `double` A scale factor for all duration values.
 * @details The device implementation reports this scale factor. A client must multiply any raw duration value received from the device
 * by this factor to obtain the physical duration. The unit of the physical duration is given by @ref QDMI_DEVICE_PROPERTY_DURATIONUNIT.
 * @note If querying this property returns @ref QDMI_ERROR_NOTSUPPORTED, a client should assume a default value of `1.0`.
 */
QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR = 13,
```

```c++
/**
 * @brief `uint64_t` The raw, unscaled duration of an operation.
 * @details To obtain the physical duration, a client must scale the raw value of this property.
 * The physical duration is calculated as: `raw_value * scale_factor`, where `scale_factor` is the value of the
 * @ref QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR property. The resulting value is in units of @ref QDMI_DEVICE_PROPERTY_DURATIONUNIT.
 * @see QDMI_DEVICE_PROPERTY_DURATIONUNIT
 * @see QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR
 */
```

Similar improvements would be due for the length-related properties and other time-related properties.

Maybe this deserves its own PR separate from the addition of the mean shuttling speed here.

include/qdmi/constants.h
```diff
@@ -616,6 +652,23 @@ enum QDMI_OPERATION_PROPERTY_T {
...
+   * duration. The value must be multiplied by the value returned by @ref
+   * QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR and divided by the value
+   * returned by @ref QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR to obtain the
+   * velocity in the unit specified by the @ref
+   * QDMI_DEVICE_PROPERTY_LENGTHUNIT divided by @ref
+   * QDMI_DEVICE_PROPERTY_DURATIONUNIT.
+   * @see QDMI_DEVICE_PROPERTY_LENGTHUNIT
+   * @see QDMI_DEVICE_PROPERTY_DURATIONUNIT
+   */
+  QDMI_OPERATION_PROPERTY_MEANSHUTTLINGSPEED = 7,
```
burgholzer: Just my two cents here: I think it pays of to be rather explicit here.
```
/**
 * @brief `uint64_t` The raw, unscaled mean shuttling speed of an operation.
 * @details To obtain the physical speed, a client must scale the raw value of this property.
 * The physical speed is calculated as: `raw_value * length_scale_factor / duration_scale_factor`.
 * The `length_scale_factor` is the value of @ref QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR and the
 * `duration_scale_factor` is the value of @ref QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR.
 * The resulting value is in units of @ref QDMI_DEVICE_PROPERTY_LENGTHUNIT per @ref QDMI_DEVICE_PROPERTY_DURATIONUNIT.
 * @note This property is mainly required for neutral atom devices where atoms representing qubits can be moved to different sites.
 * @see QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR
 * @see QDMI_DEVICE_PROPERTY_DURATIONUNIT
 * @see QDMI_DEVICE_PROPERTY_DURATIONSCALEFACTOR
 */
QDMI_OPERATION_PROPERTY_MEANSHUTTLINGSPEED = 7,
```

examples/fomac/example_fomac.cpp
```diff
@@ -166,20 +166,45 @@ auto FoMaC::get_site_id(QDMI_Site site) const -> uint64_t {
...
+    throw std::runtime_error("Unrecognized unit: " + unit);
+  }
+  return scale_factor;
+}
+
 auto FoMaC::get_site_t1(QDMI_Site site) const -> double {
-  double t1 = 0;
+  uint64_t t1 = 0;
   const int ret = QDMI_device_query_site_property(
       device, site, QDMI_SITE_PROPERTY_T1, sizeof(double), &t1, nullptr);
```
burgholzer:
```suggestion
  uint64_t t1 = 0;
  const int ret = QDMI_device_query_site_property(
      device, site, QDMI_SITE_PROPERTY_T1, sizeof(uint64_t), &t1, nullptr);
```

examples/fomac/example_fomac.cpp
```diff
@@ -166,20 +166,45 @@ auto FoMaC::get_site_id(QDMI_Site site) const -> uint64_t {
...
   throw_if_error(ret, "Failed to query the T1 time");
-  return t1;
+  return static_cast<double>(t1) * get_us_scale_factor();
 }
 
 auto FoMaC::get_site_t2(QDMI_Site site) const -> double {
-  double t2 = 0;
+  uint64_t t2 = 0;
   const int ret = QDMI_device_query_site_property(
       device, site, QDMI_SITE_PROPERTY_T2, sizeof(double), &t2, nullptr);
```
burgholzer:
```suggestion
  uint64_t t2 = 0;
  const int ret = QDMI_device_query_site_property(
      device, site, QDMI_SITE_PROPERTY_T2, sizeof(uint64_t), &t2, nullptr);
```

burgholzer (changes requested):
> @burgholzer While implementing the support for the shuttling speed (mean shuttling speed for now) as requested in https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/191, I realized that the unit of the shuttling speed is a combination of the length and duration units. The current state of this PR, hence, adds similarly to the length unit, the properties for a duration unit. As mentioned in https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/185 (https://github.com/Munich-Quantum-Software-Stack/QDMI/issues/185#issuecomment-3032103836), the introduction of the duration unit is a breaking change, as all properties related to duration values must be adapted where there semantics used a unit fixed by QDMI previously. Consequently, I tagged this PR as Minor. Let me know, whether we want to add those features now for the shuttling speed or whether we want to keep it simple for now and also just fix the unit for the shuttling speed to um/us in the QDMI docstring.

Thanks for working on this. As you'll see from the comments, I am generally happy with this direction (and the change for the existing duration quantities). However, the docstrings themselves will still need a bit of work. I think we can still target this for the 1.2 release of QDMI. Otherwise we would unnecessarily be holding off the release of this change until a v1.3. I'd rather cause more disruption now than to spread it out.

This brings me to one of the last things I noticed: None of the recent PRs added changelog entries and upgrade guide notes. This will definitely be needed, but is probably better suited as another separate PR before preparing the v1.2 release. We just must not forget this.

ystade:
> This brings me to one of the last things I noticed: None of the recent PRs added changelog entries and upgrade guide notes. This will definitely be needed, but is probably better suited as another separate PR before preparing the v1.2 release. We just must not forget this.

This is true... I would then do it directly in one single PR. Should the features be summarized in one bullet point stating something like NA support or should it list all features separateyl?

include/qdmi/constants.h
ystade: The changes for the duration unit can be incorporated into this PR, since this PR first introduces the duration unit. The changes for the length unit, I would then put into another PR.

burgholzer:
> > This brings me to one of the last things I noticed: None of the recent PRs added changelog entries and upgrade guide notes. This will definitely be needed, but is probably better suited as another separate PR before preparing the v1.2 release. We just must not forget this.
> 
> This is true... I would then do it directly in one single PR. Should the features be summarized in one bullet point stating something like NA support or should it list all features separateyl?

I would probably keep the changelog itself rather short. Potentially even just as a single bullet point that includes all the PRs and authors.
However, the upgrade guide should contain a more detailed list of features and changes. People should know what they can do now that they could not do before.
NA support might even deserve a separate page in the documentation that could substitute some of the upgrade guide.

Any breaking changes to the existing interface (thinking about the duration units here) should be listed separately in the changelog and the upgrade guide.

include/qdmi/constants.h
burgholzer: That makes sense. That follow-up PR should happen soon after this one then 😌

burgholzer (approved):
This looks and reads good to me now! Feel free to merge once happy. 

Edit: Do you see an easy way to improve the coverage here?

ystade:
@burgholzer Just saw the codecov report. These three lines are actually hard to cover. They can only be covered with a three more device implementations, which (honestly) I am not willing to implement. I see two options: either just ignoring the codecov report or removing those lines from the FoMaC.

burgholzer:
> @burgholzer Just saw the codecov report. These three lines are actually hard to cover. They can only be covered with a three more device implementations, which (honestly) I am not willing to implement. I see two options: either just ignoring the codecov report or removing those lines from the FoMaC.

Let's ignore the (patch) coverage for now. 
I agree that this does not make sense to implement at this point in time.

PR #218, opened 2025-08-28
Fix bug in example tool

rainij:
## Description

There was a bug in `Tool::compile`. The original code could never (correctly) replace the qubits (e.g. `q[0]` by `q[<replaced id>]`), due to two subtle issues

- Clearing the stringstream did not actually empty it (see e.g. `from.clear()`), hence it accumulated garbage preventing the pattern matching to ever match the qubits.
- `edge.first` and `edge.second` are just pointer and hence print to but numbers and not the desired ids of the sites.

This PR fixes those things. There is a test for `Tool::compile` (see `TEST_P(QDMIImplementationTest, ToolCompile)`) but it could not capture this bug because the input and expected output are mostly the same which is replicated by doing (mostly) nothing.

## Checklist:

      removals.
- [ ] The changes follow the project's style guidelines and introduce no new warnings.

### DIFF
--- a/examples/tool/example_tool.cpp
+++ b/examples/tool/example_tool.cpp
@@ -67,22 +67,29 @@ std::string Tool::compile(const std::string &qasm_string) {
     throw std::invalid_argument(
         "The device does not provide enough qubits for the circuit.");
   }
+
   // Choose an arbitrary edge for the two qubits
   const auto edge = fomac.get_coupling_map().front();
-  std::stringstream from;
-  from << "qreg q[" << num_qubits << "];";
-  std::stringstream to;
-  to << "qreg q[" << fomac.get_qubits_num() << "];";
-  auto result = replace_all_occurrences(qasm_string, from.str(), to.str());
-  from.clear();
-  from << "q[0]";
-  to.clear();
-  to << "q[" << edge.first << "]";
-  result = replace_all_occurrences(result, from.str(), to.str());
-  from.clear();
-  from << "q[1]";
-  to.clear();
-  to << "q[" << edge.second << "]";
-  result = replace_all_occurrences(result, from.str(), to.str());
+  auto id0 = fomac.get_site_id(edge.first);
+  auto id1 = fomac.get_site_id(edge.second);
+
+  std::string from_decl = (std::stringstream() << "qreg q[" << num_qubits << "];").str();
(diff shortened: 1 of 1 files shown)


ystade:
Hi @rainij thanks for spotting this bug. Do not care about the two currently failing tests. That is due to some permission problem for PRs that come from forks.

ystade:
Ok, I should have been more precise, the `CI / Lint` job is failing due to a reason. The previous message only applies to the `CI / Build and Deploy` and `CI / Test / Coverage` job.

rainij:
Yes I have seen that the lint fails, will look into that. The setup in this repo is a bit differnt to what I am used to.

ystade:
Definitely, thank you for your effort! I am happy to review the change once ready. Feel free to request a review once you are satisfied with your changes.

rainij:
Most of the checkboxes do not really apply here IMO. Hope it is OK to not have checked all of them. So far I did not add any test for this. It does not look easy to add anything minimal here. Manually I tested it by removing the connectivity from 0 to 1 so that the output was forced to be different and map to qubits 1 and 2. But I see no minimal way of testing this within the current setup. Any thoughts on this?

ystade:
Could the current test be tweaked in a way that it surfaces the previous bug and runs through now successfully?

If not, I think, it is fine to leave the tests as they are. Despite all, the `Tool` is just an example of how to use QDMI (and FoMaC).

rainij:
> Could the current test be tweaked in a way that it surfaces the previous bug and runs through now successfully?

I thought a while about this but I couldn't find a way which only adjusts the test itself. The problem is that the compiler tool requires exactly two qubits (0 and 1) in the input and replaces those with the first edge of the CXX device (which is 0 and 1 too).

rainij:
@ystade is now ready for review. I am not sure what the `Check` CI check wants to tell me, I have the feeling it is just a summary and just reports about the other two failures. Locally all tests succeed (`ctest --test-dir build/dev/`).

ystade:
Will try to review asap.

ystade:
✅  In a local run, this PR achieves a code coverage of 94% (unchanged to before the PR). In particular, all changed lines are covered.

ystade (approved):
Nice catch and the code looks good to me. I have nothing to complain about.

Issue #215, opened 2025-08-15
🤝 Design discussion: client interface, interoperability, and “stack-as-device” composition


burgholzer:
🤖 *AI text below* 🤖

Lead: @burgholzer.

This is a kick-off discussion to settle several architectural decisions around QDMI, interoperability, and how we expose an entire software stack to other stacks.

Context recap
- QDMI today is two C interfaces:
  - Device interface (C header/ABI): implemented by providers and third parties; each implementation is nameshifted via a unique prefix to avoid symbol collisions in-process. This interface contains device/property queries and job submission/management (plus authentication via sessions).
  - Client interface (C header/ABI): a neutral, non-nameshifted mirror of the device interface that adds a device handle as first argument to most methods. A QDMI “driver” loads one or more provider device libraries (nameshifted), exposes enumeration and returns opaque handles to use devices via the client interface.
- Above QDMI we’re building:
  - A compiler infrastructure (MLIR-based) that uses QDMI queries to tailor compilation and QDMI job APIs to submit programs and retrieve results; sessions handle auth.
  - A scheduler and other orchestration components.
  - Frontend adapters (e.g., Qiskit, PennyLane) that drive the stack.

Big question 1 — Do we need the client interface? Should we standardize the compiler↔QDMI boundary?
- Observation: While working on a QDMI driver implementation for the MQT (see https://github.com/munich-quantum-toolkit/core/pull/1010) we noticed that our C++ driver already has an object-oriented abstraction. Writing a FoMaC C++ wrapper over the C client interface risks duplicating many of the same concepts, feeling redundant.
- Tension: 
  - If we let the driver expose a FoMaC API directly (skipping the client interface), the compiler stack becomes tightly coupled to the specific driver. That improves ergonomics short-term but loses plug-and-play interoperability with other drivers/stacks.
  - If we keep the client interface as the canonical boundary, we preserve the ability to mix-and-match compilers, drivers, and devices across organizations (e.g., MQV collaborations), at the cost of boilerplate and a stricter ABI.
- Options (not mutually exclusive):
  1) Keep and evolve the client interface (recommended for interoperability):
     - Formalize versioning and capability negotiation (feature bits/extensions, error codes, profiles).
     - Reference FoMaC bindings (C++/Rust/Python) that wrap the client interface in ergonomic types while remaining driver-agnostic.
     - Provide conformance tests for drivers and bindings.
  2) Driver-native high-level API (ergonomics first):
[...]

PR #230, opened 2025-10-15
✨ Enabling Multi-Core Architectures

### LINKED ISSUE #141: Recommended procedure for submitting batches
What is the recommended procedure for submitting batch jobs?

**Option 1**: Create a `QDMI_Device_Job *job` that contains an entire batch (so that you can submit all circuits from a common batch at once)

**Option 2**: Create a `QDMI_Device_Job *job` that contains just one circuit from the batch (which would require submitting each circuit in separate submissions within a single session)

### LINKED ISSUE #227: ✨ Multicore Support

The recent development of new quantum multicore device architectures, for instance that of planqc's  “Multicore Atomic Quantum Computing Systems” (MAQCS)  (https://www.lrz.de/en/research/research-projects/detail/maqcs), underscores the necessity to introduce support for architectures with more then one processing unit within the QDMI interface. The specification needs to be sufficiently flexible to accommodate not only homogenous architectures with quantum processing units that bear similarity to the concept of cores and threads in classical computing, but also potential heterogeneous architectures in the future.

The objective of the interface would be to facilitate the ability to address each processing unit for the query, as well as the job interface. Therefore, in the case of a noise model that has been queried, a noise-aware compiled circuit may be submitted to the same processing unit.

Note: Only a subset of the issue was addressed by PR #200.

FYI: 
@echavarria-lrz 
@burgholzer 
@ystade 
@AdVetter

I can think of three ways to achieve this, which I will outline below. The first two will involve breaking changes, whereas the third will not.

1. Introduce a new opaque pointer for the abstraction of a quantum processing unit that can represent a core, thread or chiplet. This opaque pointer can replace the device pointer in the interfaces used for querying and submitting jobs.
2. Add an additional argument to the query and job submission interface functions to represent the processing unit ID.
3. Introduce a new function (e.g. 'select_qpu'), which can be used to set the active core/thread/chiplet in the device pointer. This determines which core/thread/chiplet is targeted in all subsequent calls to the query and job interfaces.

ystade:
## Description

This PR adds three properties and parameters to the interface to enable the following features of a multi-core architecture:
- A (top-level) device combines several child devices (cores, processing units, ...). In QDMI these are modelled as child devices. The top-level device can be queried for its child devices through the newly added `QDMI_DEVICE_PROPERTY_CHILDDEVICES`. The client can retrieve all relevant information about the child devices via directly querying the respective properties of the child devices, such as its status, its fidelity data, ... To facilitate correct handling of the child device handle in the QDMI driver, a new session parameter, `QDMI_DEVICE_SESSION_PARAMETER_CHILDDEVICE` was added.
- The top-level devices manages and synchronizes the job execution across all child devices. In this case, the jobs can still be created for each child device but these can only be submitted via a batch job that combines all the jobs for the child devices into a single batch via the top-level device. For this, the new program format `QDMI_PROGRAM_FORMAT_BATCHJOB` was added.

Fixes #227, #141

## Usage Example

Let's consider the following example that from the client's perspective
1. initializes a client session
2. retrieves an available device
3. retireves all child devices from that device
4. creates a batch job

Note: The following exampe might not be syntactically correct.

```mermaid
sequenceDiagram
    participant Client
    participant Driver
    participant Device
    note right of Client: Retrieve device and child devices
    Client->>Driver: QDMI_session_alloc(&session)
    Client->>Driver: QDMI_session_init(session)
    Client->>Driver: QDMI_session_query_session_property(session, QDMI_SESSION_PROPERTY_DEVICES, ..., devices.data(), nullptr)
    Client->>Client: device = devices.front()
    Client->>Driver: QDMI_device_query_device_property(device, QDMI_DEVICE_PROPERTY_CHILDDEVICES, ..., clientChildDevices, nullptr)
    Driver->>Device: QDMI_device_session_query_device_property(device_session, QDMI_DEVICE_PROPERTY_CHILDDEVICES, ..., childDevices, nullptr)
    loop for (childDevice : childDevices)
        Driver->>Device: QDMI_device_session_alloc(&deviceSession)
        Driver->>Device: QDMI_device_session_set_parameter(deviceSession, QDMI_DEVICE_SESSION_PARAMETER_CHILDDEVICE, ..., childDevice)
        Driver->>Device: QDMI_device_session_init(deviceSession)
        Driver->>Driver: wrap the initialized device session into a QDMI_Device in the same way as for the device itself
    end
    Driver-->>Client: wrappedChildDevices
    note right of Client: Create and submit batch job
    loop for (childDevice: clientChildDevices)
        Client->>Driver: QDMI_device_create_job(childDevice, &clientChildJob)
        Driver->>Device: QDMI_device_session_create_device_job(childDevice->deviceSession, &childJob)
        Driver-->>Client: wrapped child job
    end
    Client->>Driver: QDMI_device_create_job(device, &clientJob)
    Driver->>Device: QDMI_device_session_create_device_job(device->deviceSession, &job)
    Driver-->>Client: wrapped (batch) job
    Client->>Driver: QDMI_job_set_parameter(clientJob, QDMI_JOB_PARAMETER_PROGRAMFORMAT, ..., QDMI_PROGRAM_FROMAT_BATCHJOB)
    Client->>Driver: QDMI_job_set_parameter(clientJob, QDMI_JOB_PARAMETER_PROGRAM, ..., &clientChildJobs)
    Client->>Driver: QDMI_job_submit(clientJob)
    Driver->>Device: QDMI_device_job_submit(clientJob->deviceJob)
```

## Checklist:

      removals.


ystade:
This is also related to #215

ystade:
I moved the discussion from #227 to this PR to be able to combine it with actual additions to the code. Please, have a look and give feedback. Also feel free to ask any questions.

@burgholzer @mnfarooqi @echavarria-lrz @AdVetter

ystade:
@echavarria-lrz please, tag further perople in case I forgot someone.

echavarria-mqv:
> @echavarria-lrz please, tag further perople in case I forgot someone.

@jetridl @denialhaag

mnfarooqi:
Thank you @ystade for working on this and suggesting a solution.

It took me a while to understand how things could work without an example implementation, but I think I get it now. The proposal seems reasonable to me. However, we would need to expand the specification documentation for clarity.

> The top-level devices manages and synchronizes the job execution across all child devices. In this case, the jobs can still be created for each child device but these can only be submitted via a batch job that combines all the jobs for the child devices into a single batch via the top-level device. For this, the new program format `QDMI_PROGRAM_FORMAT_BATCHJOB` was added.


I cannot see this specified in the PR. However, there is no need for the specifications to entangle batch jobs and child devices. I am not sure why a job created on a child device cannot simply be submitted to that device, without the need for the parent device to submit it via a batch job.

burgholzer:
I am moving this to the 1.3.0 milestone for now as there still seems to be ongoing discussions that need to be resolved and the changes on `main` have been pretty stable for a while.

ystade:
> > The top-level devices manages and synchronizes the job execution across all child devices. In this case, the jobs can still be created for each child device but these can only be submitted via a batch job that combines all the jobs for the child devices into a single batch via the top-level device. For this, the new program format `QDMI_PROGRAM_FORMAT_BATCHJOB` was added.
> 
> I cannot see this specified in the PR. However, there is no need for the specifications to entangle batch jobs and child devices. I am not sure why a job created on a child device cannot simply be submitted to that device, without the need for the parent device to submit it via a batch job.

@mnfarooqi To come back to your point, yes, that is indeed not specified in the PR, as I also do see other possibilities than the described one, a device could handle batch jobs. In the discussed case, you cannot execute individual jobs on the sub-devices; they do need to be pooled in one batch job in the parent device. However, there might be architectures where the former is actually possible in addition to the batch jobs. I see the batch job as a rather flexible way to pool multiple jobs, might even be jobs on the very same device that are submitted as one batch. 

P.S.: And yes, I agree, it is hard to grasp w/o an example. However, I think it does not makes sense to have an example for this very specific case in the QDMI repository.

mnfarooqi:
> I am moving this to the 1.3.0 milestone for now as there still seems to be ongoing discussions that need to be resolved and the changes on `main` have been pretty stable for a while.

Could we please move forward with this so that it is included in the next release?

ystade:
From my side, this is already settled and could go in. I was awaiting a response of the involved parties whether the solution solves the issue. So if this is the case, this can go into the next release.

mnfarooqi:
@echavarria-mqv @AdVetter

Do you have any comments? Is it good to go in the next release?

AdVetter:
> @echavarria-mqv @AdVetter
> 
> Do you have any comments? Is it good to go in the next release?


Let's test with this simple approach. When someone working on this is limited by this idea we can patch it.

burgholzer (approved):
Alright. Let's get this in.
Since all changes here are additive, this is fine as a patch release and will be `v1.3.1`

burgholzer:
just released https://github.com/Munich-Quantum-Software-Stack/QDMI/releases/tag/v1.3.1

kabu-planqc:
Dear QDMI authors,

Whilst implementing the QDMI interface contract version 1.3.1, I noticed some discrepancies I would like to clarify:

1. The batch job submission workflow (as well as the QDMI template header device.h docstrings) introduce a new (QDMI device library implementation specific) type named QDMI_Device that is supposed to hold child device / core related data.
   -  First, the nomenclature in the batch job submission workflow diagram is quite confusing as it mixes two kinds of "QDMI_Device", one on the driver, the other on the device side. I think it would help other device library authors to clearly state that these are two different types.
   - A driver can query a list of QDMI_Device data for all child devices of the target machine from the device library (and attach elements to device library sessions). To be in line with all other data types exposed by the QDMI driver library C interface functions, I assume, it is also an opaque pointer that is only used by the driver lib internally. Due to this reason I would first ask you to also add a type definition to the device.h (or constants.h) QDMI template header files. Furthermore, I would propose to stick to the current nomenclature for opaque library pointer types and prepend an implementation specific prefix, e.g. add to the template:
   ```typedef struct my_QDMI_Device_impl_d* my_QDMI_Device;```


2. The QDMI 1.3.1 standard changes behavior of certain interface functions, depending whether they are called with a managing "client" job or one of the batch assigned task jobs as argument. However it does not specify the behavior of all job related interface functions. E.g. individual tasks cannot be cancelled but the managing job can be cancelled. What about requesting the job status for a managing job? What is supposed to be returned? QDMI_NOT_SUPPORTED or the minimum/maximum numeric enumerator value of all related task jobs? What about job resource management, should a call to my_QDMI_device_job_free(client_job) also release the associated batch task jobs? What about setting the number of shots on the managing job, should it propagate the value to the assigned task jobs, or just return not supported? It would be great to have a table of expected behavior for the two types of jobs for all exposed job interfaces.

ystade:
Hi @kabu-planqc ,

I'll try to answer your questions one-by-one:

1a) If I am not mistaken, it must be the same to be able to use the same query functions on the (parent) device and child devices from the client side. Otherwise, one would need to duplicate all functions for the childe device type.

1b) The template is meant to be as generic as possible to be useful for as many device implementers as possible. Hence, I would be hesitant to add a definition of specific types necessary for multi-core architecture, which I would assume is very non-standard. However, I agree that you most likely need such a type definition in your device implementation to support child devices.

2) Yeah, I think, when we designed this addition, we were assuming that every device defines its own behavior in this regard. However, maybe it is better to define more strictly in the interface which behavior is expected. @kabu-planqc since your are using this feature, do you have a prototype of this table? @burgholzer what would be your opinion in this regard?

burgholzer:
> Yeah, I think, when we designed this addition, we were assuming that every device defines its own behavior in this regard. However, maybe it is better to define more strictly in the interface which behavior is expected. @kabu-planqc since your are using this feature, do you have a prototype of this table? @burgholzer what would be your opinion in this regard?

My, very quick, opinion is that the multi-core feature is very much experimental and we tried to make the minimum changes necessary to facilitate it. I agree that it would probably be good to make some of the assumptions more explicit. In order to facilitate this, a concrete implementation/proposal would be very helpful.

kabu-planqc:
First of all, many thanks for the rapid response!

I interpreted the QDMI_Device as two separate types due to following reasons:

1. In the batch submission diagram the driver fetches the child devices from an initialized device library session.
2. It then creates one device library session for each child device and attaches one QDMI_Device child to said session. 
3. On the driver side, the created session is then wrapped into a QDMI_Device again?

```mermaid
sequenceDiagram
    participant Client
    participant Driver
    participant Device
    Driver->>Device: QDMI_device_session_query_device_property(device_session, QDMI_DEVICE_PROPERTY_CHILDDEVICES, ..., childDevices, nullptr)
    loop for (childDevice : childDevices)
        Driver->>Device: QDMI_device_session_alloc(&deviceSession)
        Driver->>Device: QDMI_device_session_set_parameter(deviceSession, QDMI_DEVICE_SESSION_PARAMETER_CHILDDEVICE, ..., childDevice)
        Driver->>Device: QDMI_device_session_init(deviceSession)
        Driver->>Driver: wrap the initialized device session into a QDMI_Device in the same way as for the device itself
    end
    Driver-->>Client: wrappedChildDevices
```

If a driver implementation is supposed to create QDMI_Device instances (and access members, as shown in the diagram) then it cannot be opaque. QDMI, however, represents pure C interfaces (at least all functions are exposed without name mangling) and, thus, if I expose a (transparent) type, then it has to be a plain old struct, right? Why complicate things and enforce a (plain old) data type that has to be shared across two library boundaries? 

> 1a) If I am not mistaken, it must be the same to be able to use the same query functions on the (parent) device and child devices from the client side. Otherwise, one would need to duplicate all functions for the childe device type.

I would argue it is the responsibility of the driver author to introduces a common type for the client query interfaces.

ystade:
There, I think, you have a valid point. So, this would mean for QDMI that `QDMI_DEVICE_SESSION_PROPERTY_CHILDDEVICES` would actually need to return the other type, let's call it `QDMI_Child_Device` for now. However, as you said, the driver must wrap these, or rather the session with these into usual `QDMI_Device` objects, correct? (The device session's function signature set_parameter would also need to be adapted to use the new type then).

kabu-planqc:
> There, I think, you have a valid point. So, this would mean for QDMI that `QDMI_DEVICE_SESSION_PROPERTY_CHILDDEVICES` would actually need to return the other type, let's call it `QDMI_Child_Device` for now. However, as you said, the driver must wrap these, or rather the session with these into usual `QDMI_Device` objects, correct? (The device session's function signature set_parameter would also need to be adapted to use the new type then).

Correct, at least that is how I imagined it whilst implementing our solution.

I introduced a new type in the device.h for now (but would feel better if a lib user could auto generate it from the QDMI template--as it reduces potential errors, such as mixed qdmi versions in drivers ready for mujltiple quantum machines): ```typedef struct my_QDMI_Device_impl_d* my_QDMI_Device;```

kabu-planqc:
Currnetly, I have implemented the job related interfaces as follows:

| QDMI Interface | managing job | task job |
| --- | --- | --- |
| PREFIX_QDMI_device_job_submit | supported (parallel submit) | QDMI_ERROR_NOTSUPPORTED |
| PREFIX_QDMI_device_job_cancel | supported (parallel cancellation) | QDMI_ERROR_NOTSUPPORTED | 
| PREFIX_QDMI_device_job_check | supported (no queued state) | supported | 
| PREFIX_QDMI_device_job_wait | supported (parallel wait for all) | supported |
| PREFIX_QDMI_device_job_free | supported (only itself) | supported |
| PREFIX_QDMI_device_job_get_results | QDMI_ERROR_NOTSUPPORTED | supported |
| get/set QDMI_DEVICE_JOB_PROPERTY_SHOTSNUM | QDMI_ERROR_NOTSUPPORTED | supported |
| get QDMI_DEVICE_JOB_PROPERTY_PROGRAM | QDMI_ERROR_NOTSUPPORTED | supported |

PREFIX_QDMI_device_job_free: One could argue that the whole batched job concept could benefit if we internally move the "ownership" of the individual batch tasks to the managing job they are attached to. This could remove the necessity to free all the individual task jobs themself, but just free the managing job. On the other hand this might introduce dangling task job pointers on the driver side the user is unaware of, and their successive usage to undefined behavior. Hence, an interface standard would be appreciated.

set QDMI_DEVICE_JOB_PROPERTY_SHOTSNUM: Consistently setting the amount of shots through the managing job might also be a nice convenience feature. 

PREFIX_QDMI_device_job_get_results on the managing job could be implemented to return a "list of results".

ystade:
@kabu-planqc This looks pretty reasonable. Would you mind opening a corresponding pull request that we can then review?

PR #252, opened 2025-11-26
✨ Add device property `QDMI_DEVICE_PROPERTY_SUPPORTEDPROGRAMFORMATS` for efficiently querying the device's supported program formats

burgholzer:
## Description

Adds a new device-level property for conveniently querying the supported program formats of a device.
Previously, this required creating a throwaway `QDMI_Job` and probing `QDMI_(device_)job_set_parameter` (which would error out on job handles that are null).

## Checklist:

      removals.

### DIFF
--- a/examples/fomac/example_fomac.hpp
+++ b/examples/fomac/example_fomac.hpp
@@ -35,7 +35,6 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include <vector>
 
 class FoMaC {
-private:
   QDMI_Device device;
 
   static auto throw_if_error(int status, const std::string &message) -> void;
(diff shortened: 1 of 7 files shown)


examples/fomac/example_fomac.hpp
```diff
@@ -35,7 +35,6 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #include <vector>
 
 class FoMaC {
-private:
```
ystade, line 38: Is that on purpose? Should the device, and all members below be publicly accessible?

ystade (approved):
Overall, lgtm, just one minor question?

examples/fomac/example_fomac.hpp
burgholzer: The default in classes is private, so the specifier here is redundant. CLion complained and highlighted that. So I simply removed it

PR #325, opened 2026-02-11
🎨 Install All Header into Device

ystade:
## Description

Only two of the three headers necessary for the device interface are prefixed. The `constants.h` is not. However, only due to this fact, every device depends on an available QDMI installation. This PR also adds the `constants.h` header to the files that are copied into the include directory of the device.

## Checklist:

      removals.


ystade:
The coverage is concerning. It probably dropped because the constants header is not taken from the include directory in the device implementation anymore. However, I do not see how to fix it. In particular, the very same header should still be used by the driver directly.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ clients compiled against a different minor or major version.
 
 ## [Unreleased]
 
+### Changed
+
+- 🚸 Export _all_ device headers (incl. `constants.h`) into device implementation ([#325]) ([\@ystade])
```
```suggestion
- 🚸 Export _all_ device headers (incl. `constants.h`) as part of the device implementation ([#325]) ([\@ystade])
```
I was wondering whether one should highlight that QDMI devices now no longer need to explicitly link against the QDMI header-only library?

burgholzer (approved):
I would assume that the template itself also needs an update, but you are probably covering this in #326 (have not checked so far).

If this is supposed to go into a `1.2.2` release of QDMI, this change also needs to be back ported to the `v1.2.x` branch.
If this should, instead, constitute a `v1.3.0` release, no back port is necessary.
(Note that we do not currently have mergifyio set up like in mqt-core, which would make this a bit easier).

burgholzer (changes requested):
Ah, the reason coverage is missing is because the QDMI library is no longer linked to the example device so the respective coverage settings are no longer propagated to the target. This needs to be fixed.

ystade:
@burgholzer I adopted the cmake config quite a bit to enable coverage for devices. Kindly have a look, whether this is all fine.

ystade:
> I would assume that the template itself also needs an update, but you are probably covering this in #326 (have not checked so far).
> 
> If this is supposed to go into a `1.2.2` release of QDMI, this change also needs to be back ported to the `v1.2.x` branch. If this should, instead, constitute a `v1.3.0` release, no back port is necessary. (Note that we do not currently have mergifyio set up like in mqt-core, which would make this a bit easier).

1. Yes, the template must also be changed. It is easier to do that in #326 because otherwise the "Build and Install" job fails due to missing header files, as this refers to the `develop` branch. Hence, it is cleaner to merg this PR and adapt the template in a subsequent PR.

2. I would be fine to let that constitute a v1.3.0 or maybe a v1.2.3 release. Not needing to link a target anymore warrants a new path/minor version in my opinion.

burgholzer (approved):
I pushed a few minor cleanups. Merging this now and creating a backport for `v1.2.x`.

PR #456, opened 2026-07-07
✨ QDMI Child Device representation in a multi-core architecture

kabu-planqc:
## Description

QDMI 1.3.1 introduced the concept of multi-core architectures (follow design discussions in #230). The [QDMI 1.3.1 interface contract](https://github.com/Munich-Quantum-Software-Stack/QDMI/releases/tag/v1.3.1) proposes to employ the driver/**client** library type [QDMI_Device](https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/develop/include/qdmi/client.h) to store child device related information on the **device** library side. As discussed in #230, this introduces unnecessary coupling/friction in-between device and driver library implementations due to following facts:. 

- As the type is exposed across the device <-> driver library boundaries it cannot be opaque:
   - So far, the QDMI device library represents pure C interfaces (functions are exposed without name mangling) and only exposes opaque data types. To stay C compliant, the type would have to be a plain old struct.
   - Furthermore, driver library authors are supposed to implement the--device library agnostic--client library type QDMI_Device. Yet, device library author may want to have a say in what kind of architecture/implementation specific information is required in said type--as it is used device library internally to "attach individual batch tasks to specific cores".
   - Moreover, different device library implementations may want to store different information. Instead of finding a common ground that works for all device implementation, the child device data type should stay implementation specific, i.e. be opaque and not exposed across the device library boundary.

## Describe the solution you'd like

Decouple the types used by the **driver** library for agnostic device representations and the type used by the **device** library to hold child device related information, by

- Introducing a new opaque pointer type QDMI_Child_Device to hold device library implementation specific child device information.

- Following the concepts of implementation specific data types in QDMI, to allow for more than one child device representation in a driver library, the type name should be extended by the (device library implementation specific) prefix.

- Exposing the type definition in one of the headers, so it is known to the driver authors--even if said author auto-generates the device library headers with the QDMI prefix template renderer. E.g. add in my_qdmi/types.h: 
```typedef struct MY_QDMI_Device_impl_d* MY_QDMI_Device;```

Fixes #230  
@ystade @burgholzer <- tagging the autors, at it is my first contribution to QDMI.

## Checklist:
removals.
- [ ] The changes follow the project's style guidelines and introduce no new warnings.

### CODE REVIEW

File: cmake/prefix_defs.txt [hunk 1]
```diff
@@ -31,3 +31,5 @@ QDMI_device_EXPORTS
 QDMI_DEPRECATED
 QDMI_DEPRECATED_EXPORT
 QDMI_DEPRECATED_NO_EXPORT
+QDMI_Child_Device
+QDMI_Child_Device_impl_d
```
ystade:
```suggestion
QDMI_Child_Device
QDMI_Child_Device_impl_d
QDMI_STATIC_DEFINE
QDMI_EXPORT
QDMI_NO_EXPORT
QDMI_device_EXPORTS
QDMI_DEPRECATED
QDMI_DEPRECATED_EXPORT
QDMI_DEPRECATED_NO_EXPORT
```
I would just slightly reorder the entries.

include/qdmi/types.h
```diff
@@ -72,6 +72,24 @@ typedef struct QDMI_Site_impl_d *QDMI_Site;
...
+ * Each implementation of the @ref device_interface "QDMI Device Interface"
+ * defines the actual implementation of the concept.
+ *
+ * A simple example of an implementation is a struct that merely contains an
+ * index, which can be used to identify the respective core / processing unit.
+ * @code{.cpp}
+ * struct QDMI_Child_Device_impl_d {
+ *   size_t id;
+ * };
+ * @endcode
```
ystade:
```suggestion
 * @endcode
 * @see QDMI_DEVICE_PROPERTY_CHILDDEVICES 
 * @see QDMI_DEVICE_SESSION_PARAMETER_CHILDDEVICE 
```
I would just add some cross-references here.

ystade: As the client should not see this type, this should be placed in the `device.h` file in the same way as the `QDMI_Device_Job` type. Also see the main comment on this PR.

ystade (commented):
@kabu-planqc Thanks for this proposal. The changes look concise and clean. I just left two comments inline, one related to the following.

With these changes, it is completely clear to me how a device implementation implements the multi-core feature. However, I keep thinking about how a driver deals with that. I am still confident that, after the changes, this is feasible but not straightforward. Hence, I will briefly outline how I imagine the workflow, and @kabu-planqc would like to hear your opinion on whether this makes sense, as you already have some experience in implementing and using this feature of QDMI.

The client only sees `QDMI_Device` handles, as the client has no way to use `QDMI_Child_Device` handles in any function. Hence, in the driver, the `QDMI_Child_Device` handles must be wrapped into corresponding `QDMI_Device` handles for the client. The tricky part here is that each `QDMI_Child_Device` must have exactly one corresponding `QDMI_Device` handle on the client side. Hence, when a device library returns the same `QDMI_Child_Device` handle twice, e.g., in subsequent function calls, it must map to the same `QDMI_Device` handle. To this end, the driver must maintain some mapping from `QDMI_Child_Device` handles to `QDMI_Device` handles.

Aside from the more abstract concern above, everything looks fine to me, but I leave the final approval to @burgholzer.

kabu-planqc:
@ystade Thanks for the rapid feedback!

> The client only sees `QDMI_Device` handles, as the client has no way to use `QDMI_Child_Device` handles in any function. Hence, in the driver, the `QDMI_Child_Device` handles must be wrapped into corresponding `QDMI_Device` handles for the client. The tricky part here is that each `QDMI_Child_Device` must have exactly one corresponding `QDMI_Device` handle on the client side. Hence, when a device library returns the same `QDMI_Child_Device` handle twice, e.g., in subsequent function calls, it must map to the same `QDMI_Device` handle. To this end, the driver must maintain some mapping from `QDMI_Child_Device` handles to `QDMI_Device` handles.

Correct, as mentioned in #230, I would argue  it is the responsibility of the driver author to introduces a common type for the client query interfaces. As long as the device library guarantees that the list returned by querying QDMI_DEVICE_PROPERTY_CHILDDEVICES is in order, i.e.  the n-th element always corresponds to the n-th core of the device, defining such a mapping should be trivial.

Maybe we should add this "order" requirement to the docs of the QDMI_DEVICE_PROPERTY_CHILDDEVICES. Furthermore, we may also want to state, that defining such a mapping on the basis of the address the opaque QDMI_Child_Device references should be avoided, as the child device data may be (re-)constructed, e.g. if queried from different device session.

Lastly, according to the batch submission workflow diagram in #230, the driver wraps a (child) device_**session** into a  QDMI_Device. Hence, I am not convinced that the driver requires a mapping from QDMI_Child_Device to QDMI_Device, but rather a mapping from (child_)device_session to QDMI_Device--else the relation to one specific device session (of potentially many) created for the core gets lost.

ystade:
> Correct, as mentioned in #230, I would argue it is the responsibility of the driver author to introduces a common type for the client query interfaces. As long as the device library guarantees that the list returned by querying QDMI_DEVICE_PROPERTY_CHILDDEVICES is in order, i.e. the n-th element always corresponds to the n-th core of the device, defining such a mapping should be trivial.
> 
> Maybe we should add this "order" requirement to the docs of the QDMI_DEVICE_PROPERTY_CHILDDEVICES. Furthermore, we may also want to state, that defining such a mapping on the basis of the address the opaque QDMI_Child_Device references should be avoided, as the child device data may be (re-)constructed, e.g. if queried from different device session.
> 
> Lastly, according to the batch submission workflow diagram in #230, the driver wraps a (child) device_**session** into a QDMI_Device. Hence, I am not convinced that the driver requires a mapping from QDMI_Child_Device to QDMI_Device, but rather a mapping from (child_)device_session to QDMI_Device--else the relation to one specific device session (of potentially many) created for the core gets lost.

@kabu-planqc Thanks for the insights. This makes absolutely sense and I forgot that in between the driver must create a device session, so, indeed, the issue should not even be an issue.

ystade:
@kabu-planqc something I forgot in my review and @burgholzer reminded me of: Could also please add corresponding entries to the CHANGELOG.md and UPGRADE.md?

cmake/prefix_defs.txt
```diff
@@ -31,3 +31,12 @@ QDMI_device_EXPORTS
...
+QDMI_STATIC_DEFINE
+QDMI_EXPORT
+QDMI_NO_EXPORT
+QDMI_device_EXPORTS
+QDMI_DEPRECATED
+QDMI_DEPRECATED_EXPORT
+QDMI_DEPRECATED_NO_EXPORT
```
burgholzer, line 35: Something went wrong here. This now contains duplicate entries.

include/qdmi/device.h
```diff
@@ -716,6 +716,26 @@ QDMI_EXPORT void QDMI_device_job_free(QDMI_Device_Job job);
 
 /** @} */ // end of device_job_interface
 
+/**
+ * @brief A handle for a child device.
+ * @details An opaque pointer to an implementation of the QDMI child device
+ * concept. A child device generally represents a core or processing unit
+ * of a multi-core device.
+ * Each implementation of the @ref device_interface "QDMI Device Interface"
+ * defines the actual implementation of the concept.
```
burgholzer: Should this emphasize that only implementers that actually use the type need to implement it? 
The way this is currently framed makes it seem like every QDMI Device author needs to implement this struct. This would make this a breaking change, which would be unfortunate.

burgholzer (changes requested):
Thanks @kabu-planqc for the PR! 🙏🏼 
I only have two small inline requests left for this.

In addition, as already outlined by @ystade, please add a changelog entry for this, as well as a small section to the upgrade guide that describes how implementers should adapt.

include/qdmi/device.h
kabu-planqc: fair point, will add a note.

cmake/prefix_defs.txt
kabu-planqc: My bad, github duplicated one entry whilst auto-resolving the issue/comment. I made it worse when removing that duplicate entry, apologies. Fixed [here](https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/456/changes/9a79079960c481556ccc46333b1b9be6ea790a94).

include/qdmi/device.h
```diff
@@ -716,6 +716,30 @@ QDMI_EXPORT void QDMI_device_job_free(QDMI_Device_Job job);
...
+ * @brief A handle for a child device.
+ * @details An opaque pointer to an implementation of the QDMI child device
+ * concept. A child device generally represents a core or processing unit
+ * of a multicore device.
+ * Each implementation of the @ref device_interface "QDMI Device Interface"
+ * defines the actual implementation of the concept.
+ * Let us note here, that it is not mandatory to implement the QDMI child device
+ * concept. Only authors of a multicore device library that want to facilitate
+ * job execution on a dedicated core and/or need to expose device properties on
+ * a child device level are required to implement the concept.
```
burgholzer:
```suggestion
 * Each implementation of the @ref device_interface "QDMI Device Interface"
 * may define the actual implementation of the concept.
 * 
 * @note Only authors of a multicore device library that want to facilitate 
 * job execution on a dedicated core and/or need to expose device properties on
 * a child device level must implement the concept.
```
Small refinement

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ clients compiled against a different minor or major version.
 
 - ⬆️ Update [munich-quantum-toolkit/workflows] to version `v2.0.1` ([#426]) ([\@denialhaag])
 
+### Changed
+
+- ✨ Introduce new **device library** type `QDMI_Child_Device` intended to store core / processing-unit related data in a multicore architecture. In contrast to the proposed specifications in [#230] where the **client library** related type `QDMI_Device` is used to store such information, this new opaque type is introduced to avoid unnecessary coupling across library boundaries. ([#456]) ([\@kabu-planqc])
```
```suggestion
- ✨ Introduce new **device library** type `QDMI_Child_Device` intended to store core / processing-unit related data in a multicore architecture ([#456]) ([\@kabu-planqc])
```
We typically keep the changelog entries more concise and add more prose to the upgrade guide.

burgholzer (approved):
Thanks again for the changes @kabu-planqc 🙏🏼 
I just have two wording suggestions left, then this can go in.

We'll do one more maintenance pass on the QDMI repo itself and then release 1.3.2 with this change included.

kabu-planqc:
@ystade @burgholzer I would like to thank you for your support  & rapid responses. Collaboration was a pleasure!
Looking forward to QDMI v1.3.2...

denialhaag:
We have just released `v1.3.2`. 🙂

PR #485, opened 2026-08-05
✨ Add support for retrieving existing jobs by ID

burgholzer:
🤖 *AI text below* 🤖

## Description

Adds retrieval of existing jobs by ID for the next patch release (1.3.3). Clients can obtain a fresh local handle for a previously submitted job from its QDMI job ID. Retrieval uses the current session credentials and does not clone, submit, or modify the remote job.

The device interface and generated prefix definitions expose the corresponding provider symbol. The example device returns `QDMI_ERROR_NOTSUPPORTED`, while the generated device template provides the symbol and defaults to `QDMI_ERROR_NOTIMPLEMENTED`. Drivers remain ABI-compatible with provider libraries built against older QDMI releases: an absent symbol is treated as unsupported.

Freeing a local job handle has device-specific effects on the underlying job; the lifetime semantics are documented on the client and provider free functions.

## Validation

- Debug build and all 103 CTest tests passed, with the expected read-only skips.
- Generated-device symbol checks and template builds passed.
- `uvx prek run --all-files` passed.
- `git diff --check` passed.


include/qdmi/device.h
```diff
@@ -478,6 +482,46 @@ QDMI_EXPORT int
...
+ * @return @ref QDMI_ERROR_NOTFOUND if no accessible job with @p job_id exists.
+ * @return @ref QDMI_ERROR_BADSTATE if @p session is not initialized.
+ * @return @ref QDMI_ERROR_PERMISSIONDENIED if @p session is not permitted to
+ * access the job.
+ * @return @ref QDMI_ERROR_FATAL if opening the job failed due to a fatal
+ * error.
+ *
+ * @attention Drivers must continue loading device implementations built
+ * against QDMI versions that predate this function and report
+ * @ref QDMI_ERROR_NOTSUPPORTED when opening a job is requested.
```
burgholzer:
```suggestion
 * error.
```

templates/device/src/my_device.cpp
```diff
@@ -92,6 +92,12 @@ int MY_QDMI_device_session_create_device_job(MY_QDMI_Device_Session session,
   return QDMI_ERROR_NOTIMPLEMENTED;
 }
 
+int MY_QDMI_device_session_open_device_job(MY_QDMI_Device_Session session,
+                                           const char *job_id,
+                                           MY_QDMI_Device_Job *job) {
+  return QDMI_ERROR_NOTSUPPORTED;
```
burgholzer:
```suggestion
  return QDMI_ERROR_NOTIMPLEMENTED;
```

burgholzer (commented):
Two small corrections.

include/qdmi/client.h
```diff
@@ -629,6 +632,38 @@ typedef struct QDMI_Job_impl_d *QDMI_Job;
...
 
+/**
+ * @brief Open an existing job.
+ * @details Creates a new local job handle for the existing remote job
+ * identified by @p job_id. Opening a job does not submit, clone, or otherwise
+ * modify the remote job. The returned handle can be used to query properties,
+ * check or wait for completion, cancel the job, and retrieve results.
+ *
+ * The job is accessed with the credentials and configuration of the current
+ * session. The job ID is an opaque identifier, not an authentication
```
ystade: I find it confusing to use the adjective "opaque" here because, even though the ID may be cryptic, I can indeed read it, it is just a string. I think just dropping the word "opaque" is fine.
```suggestion
 * session. The job ID is an identifier, not an authentication
```

include/qdmi/client.h
```diff
@@ -629,6 +632,38 @@ typedef struct QDMI_Job_impl_d *QDMI_Job;
...
+ * credential. Parameters cannot be set on an opened job, and an opened job
+ * cannot be submitted again.
```
ystade: Why not? I could imagine, one configures a job, but for some reason does not submit it and only later logs in again and wants to submit the job.

include/qdmi/device.h
```diff
@@ -478,6 +482,42 @@ QDMI_EXPORT int
...
 
+/**
+ * @brief Open an existing device job.
+ * @details Creates a new local device-job handle for the existing remote job
+ * identified by @p job_id. Opening a job does not submit, clone, or otherwise
+ * modify the remote job. The returned handle can be used to query properties,
+ * check or wait for completion, cancel the job, and retrieve results.
+ *
+ * The job is accessed with the credentials and configuration of @p session.
+ * The job ID is an opaque identifier, not an authentication credential.
```
ystade:
```suggestion
 * The job ID is an identifier, not an authentication credential.
```
See above.

include/qdmi/device.h
```diff
@@ -478,6 +482,42 @@ QDMI_EXPORT int
...
+ * Parameters cannot be set on an opened job, and an opened job cannot be
+ * submitted again.
```
ystade: Why not? See above.

include/qdmi/client.h
```diff
@@ -629,6 +632,38 @@ typedef struct QDMI_Job_impl_d *QDMI_Job;
...
+ * is @c NULL, or if @p job_id is empty.
+ * @return @ref QDMI_ERROR_NOTSUPPORTED if the driver or device does not
+ * support opening existing jobs.
+ * @return @ref QDMI_ERROR_NOTFOUND if no accessible job with @p job_id exists.
+ * @return @ref QDMI_ERROR_PERMISSIONDENIED if the current session is not
+ * permitted to access the job.
+ * @return @ref QDMI_ERROR_FATAL if opening the job failed due to a fatal
+ * error.
+ */
+int QDMI_device_open_job(QDMI_Device device, const char *job_id, QDMI_Job *job);
```
ystade: This kind of function also exists in V2 but under another name, see https://github.com/Munich-Quantum-Software-Stack/QDMI/blob/cbd06aba9d4d4dcb9e0969e7db23418c7ccdd9db/include/qdmi/job/functions.h#L306

I actually like the other name more because it is more specific, and the user knows immediately what the function is doing and which parameters it needs. If changed, these changes must also be reflected in the docstrings.
```suggestion
int QDMI_session_retrieve_job_by_id(QDMI_Device device, const char *job_id, QDMI_Job *job);
```

include/qdmi/device.h
```diff
@@ -478,6 +482,42 @@ QDMI_EXPORT int
...
+ *
+ * @param[in] session The initialized session with which to open the job. Must
+ * not be @c NULL.
+ * @param[in] job_id The nonempty, null-terminated ID returned by
+ * @ref QDMI_DEVICE_JOB_PROPERTY_ID. Must not be @c NULL.
+ * @param[out] job A pointer to a handle that will store the opened job. Must
+ * not be @c NULL. The handle must be freed by calling @ref
+ * QDMI_device_job_free when it is no longer used. Freeing the handle does not
+ * cancel or delete the remote job.
```
ystade: Very minor: Not sure whether I would mention this here so explicitly. Some device implementations may indeed delete a job when it is freed. These device implementations obviously do not support retrieving the job by id again.

ystade (dismissed):
@burgholzer Definitely a functionailty that is needed (and also implemented in QDMIv2). Overall, the modifications look good to me. I just have a comment on naming, see the comments.

include/qdmi/client.h
burgholzer: Yeah, works for me without the "opaque" 👍🏼

include/qdmi/client.h
burgholzer: In many (if not practically all) scenarios, only the submission of a job will actually give it a persistent ID that can be used for retrieval. (Braket's Task ARN, IQM's Job UUID, etc.)
Such a persistent identifier is needed so that a job can be reconstructed across process boundaries, driver restarts, etc.

If it were possible to reconstruct a job and submit it (again), I have no idea how any provider with their own job database would handle this.

If we were to lift this restriction, it would mandate every QDMI device to manage some kind of storage of configured programs, options, etc.
Does not seem feasible to me. Would you agree?

include/qdmi/client.h
burgholzer: I like the other name. Will incorporate that consistently in the next revision!
Thanks for the suggestion!

include/qdmi/client.h
ystade: To me it seems feasible, but your argument is still reasonable, and you have convinced me that submission of retrieved jobs is not needed/does not make sense. I'll resolve.

include/qdmi/device.h
burgholzer: Hm. I see the point, which is fairly nuanced.
Devices that want to support retrieval (and implement the new function) must not delete jobs upon freeing from QDMI.
Devices that do not support persistent jobs can do what they want.
Any idea for how to make the wording better here? I can also ask Codex on this.

ystade: What about 
```suggestion
 * necessarily cancel or delete the remote job.
```
?

burgholzer: Yeah. That would work, but now that I think about it a little more deeply, this sentence (no matter how it is worded) does not even really belong here. If at all, it belongs to the `QDMI_device_job_free` method to describe its semantics. I'll revisit this in my revision.

include/qdmi/client.h
burgholzer: Done in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/485/commits/a884facd38a551086eef00c3116f1c8e9a09b9c2

include/qdmi/device.h
burgholzer: Done in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/485/commits/a884facd38a551086eef00c3116f1c8e9a09b9c2

PR #508, opened 2026-08-22
✨ Add exact payload execution capability metadata

simon1hofmann:
🤖 *AI text below* 🤖

Design leads: @burgholzer and @simon1hofmann.

## Description

Prototype payload-specific execution capabilities. This is a **non-blocking QDMI 1.4 candidate**, not a settled contract or a prerequisite for the other workstreams.

**Design gate:** #523 must record the format/encoding, supported/native operations, classical capabilities, execution guarantees and opaque-program decisions before this PR is merge-ready. Core requirements and compiler/SDK consumers are tracked in munich-quantum-toolkit/core#2365; provider counterexamples are linked as subissues of #523.

This PR remains independently based on `v1.4`. Multi-program jobs moved to #509 using the existing enum. Replaceable drivers (#511) and metadata removal (#512 → #513) do not depend on this prototype.

Specifically, this pull request:

- replaces the legacy program-format enum at the device and job boundaries with a fixed C-compatible descriptor containing the exact format ID, version, profile, and text or binary encoding;
- adds a descriptor-specific two-call feature query with fixed string feature records, optional numeric values, and typed constraints;
- defines `QDMI_ERROR_NOTSUPPORTED` as unknown optional-feature metadata, while a successful query returns the complete optional set, which may be empty;
- defines exact OpenQASM and QIR descriptors and their normative baseline semantics;
- treats QPY and IQM JSON as provider-defined formats that require vendor-namespaced descriptor IDs;
- defines flat payload-declared bit ordering for shots and histogram keys and adds format-native program output for QIR;
- removes calibration and batch pseudo-formats from the payload vocabulary because they are administration and multi-program job concerns, not program formats; and
- updates the example device, device template, conformance tests, documentation, changelog, and upgrade guide.

The fixed descriptor and feature records have a direct migration path to QDMI v2. Devices list each exact accepted descriptor; clients do not infer version compatibility.

No additional dependencies are required.

## Previous prototype validation

The checks below describe the existing implementation, not a new validation run or approval of its design.

- Release build: passed.
- Release CTest suite: 104/104 passed; read-only result tests are configured skips.
- Clang 22 lint-preset build and changed-line Clang-Tidy checks: passed; only unchanged baseline warnings remain.
- Strict C11 and C++20 public-ABI compile checks: passed.
- Embedded QIR validation with LLVM 22: passed.
- Doxygen documentation build: passed, with only the existing obsolete `DOT_MULTI_TARGETS` warning.
- `uvx prek run --all-files`: passed.
- `git diff --check`: passed.

AI assistance: OpenAI Codex assisted with implementation, review, testing, and this description.

## Contract review requirements

Review the prototype against #523's device-and-format-specific capabilities, bounded OpenQASM/QIR feature vocabulary, custom-format support, and variable compilation boundaries. Supported/native operation levels and provider-neutral execution guarantees remain design decisions. Opaque pulse transport does not imply that QDMI already supplies all compiler-construction metadata.

The capability model is a priority for 1.4, not a release blocker. Before the interface change merges, link a working Core consumer and a demonstration with at least one existing provider, preferably both where applicable. Compatibility evidence suffices when a provider needs no changes. Development revisions may support these tests; published artifacts must use released dependencies.

## Checklist

- [x] I have added migration instructions to the upgrade guide.
- [x] The changes are fully tested locally; GitHub CI will validate the supported configurations.

- [x] I have disclosed the use of AI tools in the PR description as per the [AI Usage Guidelines](https://munich-quantum-software-stack.github.io/QDMI/latest/md_docs_2ai__usage.html).
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.


examples/driver/qdmi_example_driver.cpp
```diff
@@ -230,6 +233,7 @@ void QDMI_library_load(const std::string &lib_name, const std::string &prefix) {
     LOAD_SYMBOL(library, prefix, device_job_get_results)
     // device query interface
     LOAD_SYMBOL(library, prefix, device_session_query_device_property)
+    LOAD_SYMBOL(library, prefix, device_session_query_program_features)
```
burgholzer, line 228: If we are making this unconditional here, we may as well drop the `OPTIONAL` nature of `device_session_retrieve_device_job_by_id` for this minor release.
This is only the example driver anyway.

include/qdmi/constants.h
```diff
@@ -901,208 +892,125 @@ enum QDMI_JOB_STATUS_T {
...
+
 /**
- * @brief Enum of formats that can be submitted to the device.
+ * @brief Exact program format accepted by a device.
+ * @details All fields take part in identity. The `id` and `profile` arrays must
+ * be NUL-terminated, and every byte after the first NUL must be zero. An empty
+ * profile identifies a format without a named profile. Devices list every
+ * accepted descriptor separately; clients must not infer version compatibility.
+ *
+ * Standard IDs are `openqasm`, `qir`, `qpy`, and `iqm-json`. QIR 2.1 uses
```
burgholzer: I'd almost argue that we should drop the explicit IQM JSON program format given its vendor-proprietary nature. This can be handled as a custom format equally well and also gives more freedom to QDMI-on-IQM to decide how it wants to handle its non-standard/proprietary payload formats beyond QIR.

include/qdmi/constants.h
```diff
@@ -901,208 +892,125 @@ enum QDMI_JOB_STATUS_T {
...
+ * profiles `base` and `adaptive`. OpenQASM 2.0 and 3.0 use an empty profile.
+ * QIR 2.1 is defined by the QIR specification release 2.1. Its Adaptive
+ * baseline includes mid-circuit measurement, measured-qubit reuse,
+ * measurement-result use, forward branching, and the necessary i1
+ * computation. Devices report the remaining QIR module flags as optional
+ * features. Other standardized descriptors have an empty runtime-feature
+ * baseline.
```
burgholzer: This feels like it loses a lot of the information from the previous docstrings about the matching of device capabilities to program payloads.

include/qdmi/constants.h
```diff
@@ -901,208 +892,125 @@ enum QDMI_JOB_STATUS_T {
...
+
+/**
+ * @brief One optional runtime feature accepted for an exact program format.
+ * @details `id` must be NUL-terminated, and every byte after the first NUL must
+ * be zero. Boolean features use value zero. Width-dependent features repeat
+ * the same ID with one exact width per record. Unknown feature IDs are ignored
+ * by clients that do not understand them. A reported feature means unrestricted
+ * support for that feature. A device with a nesting, trip-count, case-count, or
+ * instruction restriction must not report the feature unless the format's
+ * normative baseline already defines that restriction. Duplicate records are
```
burgholzer: > A device with a nesting, trip-count, case-count, or
> instruction restriction must not report the feature unless the format's
> normative baseline already defines that restriction.

This does not feel quite right. We may want to prepare for a future, where device libraries can also indicate such restrictions explicitly. Thus, the wording here probably needs to be "softer".

include/qdmi/constants.h
```diff
@@ -901,208 +892,125 @@ enum QDMI_JOB_STATUS_T {
...
+#define QDMI_PROGRAM_FEATURE_IR_FUNCTIONS "ir-functions"
+/// Multiple return point support.
+#define QDMI_PROGRAM_FEATURE_MULTIPLE_RETURN_POINTS "multiple-return-points"
+/// Dynamic qubit management support.
+#define QDMI_PROGRAM_FEATURE_DYNAMIC_QUBIT_MANAGEMENT "dynamic-qubit-management"
+/// Dynamic result management support.
+#define QDMI_PROGRAM_FEATURE_DYNAMIC_RESULT_MANAGEMENT                         \
+  "dynamic-result-management"
+/// Array support.
+#define QDMI_PROGRAM_FEATURE_ARRAYS "arrays"
```
burgholzer: Would it make sense to expand the docstrings of all of these definitions to give more context of what is meant with each of these features?

include/qdmi/constants.h
```diff
@@ -1166,6 +1074,15 @@ enum QDMI_JOB_RESULT_T {
    *      QDMI_JOB_RESULT_PROBABILITIES_SPARSE_KEYS
    */
   QDMI_JOB_RESULT_PROBABILITIES_SPARSE_VALUES = 8,
+  /**
+   * @brief `char*` The complete output in the native result representation of
+   * the submitted program format.
+   * @details The exact descriptor defines the byte representation. For QIR
+   * 2.1 this is the complete NUL-terminated output-schema stream, including
```
burgholzer: One does not need to be overly specific about QIR _2.1_ throughout the PR. It is sufficient to refer to this as _QIR_ (without the major and minor versions). Only mention the major version where it is genuinely used to distinguish between QIR 1 and 2 (e.g., for anything related to opaque pointers).

CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ clients compiled against a different minor or major version.
 
 ## [Unreleased]
 
+### Added
+
+- 💥 Replace program-format enums with exact format descriptors and add
+  descriptor-scoped execution-feature queries for QDMI 1.4 ([#508])
```
```suggestion
  descriptor-scoped execution-feature queries ([#508])
```
No need to explicitly mention QDMI v1.4 in the changelog that will be part of that exact release.

UPGRADING.md
```diff
@@ -7,6 +7,41 @@ releases, please refer to the
...
+means that feature metadata is unknown. Boolean features use value zero. Width
+features, such as integer computation, use one record per supported width.
+
+Device plugins and clients must use matching minor-version headers. A QDMI 1.3
+client must not pass an enum value to a QDMI 1.4 device, and a QDMI 1.4 client
+must not call the new query on a QDMI 1.3 device.
+
+Calibration and batch submission are no longer program formats. Use a device
+extension for calibration. QDMI's existing multi-program job model owns batch
+submission.
```
burgholzer: I think this needs to be a little more elaborate on the calibration part; what is a "device extension" in that regard?

> existing multi-program job model

I am not quite sure we established that? Either this still requires a change in this PR to add that multi-program model or this needs to be rewritten.

burgholzer (changes requested):
Thanks @simon1hofmann for kickstarting this.
I already pushed a couple of changes here to drive this further.
Furthermore, I collected some feedback in the comments as part of this review. I'll hand these to my Codex session now to iterate on them.

include/qdmi/constants.h
burgholzer: 🤖 *AI text below* 🤖

Implemented. IQM JSON is no longer a standard format ID. QDMI-on-IQM can expose its proprietary payloads through namespaced provider descriptors.

include/qdmi/constants.h
burgholzer: 🤖 *AI text below* 🤖

Restored and tightened the portable mappings. The descriptor contract now defines regular-site ordering, zone exclusion, qubit mapping, operation name and arity matching, QIR static indices, QPY circuit cardinality, and the baseline primitives that do not need operation records.

include/qdmi/constants.h
burgholzer: 🤖 *AI text below* 🤖

Implemented typed restrictions now. Feature records can carry conjunctive constraints; QDMI initially defines maximum control-flow nesting depth, loop iteration count, and multiway case count. Unknown or malformed constraints fail closed, while `QDMI_PROGRAM_FEATURE_UNCONSTRAINED` keeps unrestricted features simple.

CHANGELOG.md
burgholzer: 🤖 *AI text below* 🤖

Applied the suggested release-neutral wording. The entry now starts with descriptor-scoped execution-feature queries and also names the program-output result added by the same contract.

include/qdmi/constants.h
burgholzer: 🤖 *AI text below* 🤖

Expanded every standard feature and constraint definition with its execution semantics, value meaning, dependencies, and counting rule where applicable.

include/qdmi/constants.h
burgholzer: 🤖 *AI text below* 🤖

Adjusted the wording to use QIR without a version where the rule is generic. Exact versions remain only where they distinguish the legacy QIR 1.0 migration, the selected QIR specification release, or the independent output-schema version.

UPGRADING.md
burgholzer: 🤖 *AI text below* 🤖

Clarified both points. `QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION` remains, but QDMI v1 defines no portable calibration trigger. QDMI v1 also has no portable multi-program submission API, so this pull request only removes the batch pseudo-format; a real multi-program model remains separate.

burgholzer (commented):
Alright. This is looking pretty good to me already and I see no bigger blockers. Let's see how the downstream integration goes.

burgholzer (approved):
Formally approving this for now to unblock the stack.

PR #509, opened 2026-08-23
✨ Add multi-program jobs

burgholzer:
🤖 *AI text below* 🤖

## Description

Introduce native, ordered multi-program jobs independently of #508. This PR is based directly on `v1.4` and retains the existing `QDMI_Program_Format` enum. No payload-capability descriptor is required.

- Client and device `job_set_programs` atomically copy a list in one format, including the count-one path. The legacy `PROGRAM` setter is removed so both paths share the same contract.
- One shot count, job ID, status, cancellation, wait, and failure state apply to the aggregate job.
- Result index `i` corresponds to input program `i`; execution order is unspecified. `PROGRAMSNUM` also applies to retrieved jobs.
- Preserve binary payloads byte-for-byte, validate text framing, and preserve calibration submission and optional shots.
- Providers without an aggregate lifecycle/results contract may reject multi-program submission; concurrent single-program jobs remain a separate fallback.

No per-program formats, options, lifecycle, partial results, or execution-capability model are introduced. Core adoption is tracked in munich-quantum-toolkit/core#2359, with low-level extraction in munich-quantum-toolkit/core#2362. Target: QDMI 1.4 / Core 4.1; release artifacts require released dependency pins.

## Validation

Independent release build (including device templates) and all 109 configured CTests passed: 89 passed and 20 expected read-only skips. `uvx prek run -a` passed. Hosted CI must be evaluated on the new head.

## End-to-end acceptance

Before the interface change merges, link a working Core consumer and a demonstration with at least one existing provider, preferably both where applicable. Compatibility evidence suffices when a provider needs no changes. Development revisions may support these tests; published artifacts must use released dependencies. Validate aggregate identity, indexed results, cancellation, failure, and retrieval. A provider SDK's batch helper may only submit independent tasks; do not infer the aggregate contract from that helper alone.

## Checklist

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://munich-quantum-software-stack.github.io/QDMI/latest/md_docs_2ai__usage.html).
- [ ] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.


include/qdmi/constants.h
```diff
@@ -259,6 +265,10 @@ enum QDMI_DEVICE_JOB_PROPERTY_T {
   QDMI_DEVICE_JOB_PROPERTY_ID = 0,
   /**
    * @brief @ref QDMI_Program_Format The format of the program to be executed.
+   * @details A query returns @ref QDMI_ERROR_BADSTATE until a format is set.
+   * This property returns the format set through @ref
+   * QDMI_DEVICE_JOB_PARAMETER_PROGRAMFORMAT or @ref
+   * QDMI_device_job_set_programs.
```
burgholzer, line 234: Does it really make sense to maintain two ways of setting a single program payload? Wouldn't it make more sense to remove the `PROGRAM` job parameter and require everything to go through the new API? That should be rather easily possible.

burgholzer (commented):
This one is looking fine as well with the exception of one point that may need further addressing.

include/qdmi/constants.h
burgholzer: 🤖 *AI text below* 🤖

Removed both Client and Device PROGRAM parameters. Single-program jobs now use the program-list API with count one, and the duplicate example-device setter path is gone. The tests and tutorial use the same path. Release build, all 111 tests, and full lint pass. The #509 changelog entries now also credit @ystade for the QDMI v2 work.

PR #512, opened 2026-08-24
💥 Remove the calibration advisory property

burgholzer:
🤖 *AI text below* 🤖

## Description

Remove the unused `QDMI_DEVICE_PROPERTY_NEEDSCALIBRATION` advisory and close the regular property-enum gap. Preserve `QDMI_DEVICE_STATUS_CALIBRATION` and calibration-job submission. Provider-specific calibration readiness can use a custom property.

This is the first independent metadata-cleanup layer, based directly on `v1.4`; #513 follows it. It does not include payload capabilities, multi-program jobs, or replaceable drivers. The breaking development headers are identified as 1.4.0-dev.

## Validation

- Release configure/build: passed, including example and template targets.
- CTest: 83 passed; 18 expected read-only job tests skipped; no failures.
- `uvx prek run -a`: passed.
- Rebased commit signed and verified. Fresh hosted CI is pending; earlier-stack results are not evidence for this head.

AI assistance: Codex separated the stack, adapted the cleanup to develop, and ran local validation.

## Consumer compatibility

Removing the advisory does not remove calibration-job submission or the CALIBRATION status. Keep provider-specific calibration readiness and workflows in documented extensions; no common calibration workflow is standardized here.

Before the interface change merges, link a working Core consumer and a demonstration with at least one existing provider, preferably both where applicable. Compatibility evidence suffices when a provider needs no changes. Development revisions may support these tests; published artifacts must use released dependencies. Core adoption: [#2233](https://github.com/munich-quantum-toolkit/core/pull/2233).

## Checklist

- [x] The pull request only contains commits that are focused and relevant to this change.
- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [x] I have added migration instructions to the upgrade guide (if needed).
- [x] The changes follow the project's style guidelines and introduce no new warnings.
- [ ] The changes are fully tested and pass the CI checks.
- [x] I have reviewed my own code changes.

**If PR contains AI-assisted content:**

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://munich-quantum-software-stack.github.io/QDMI/latest/md_docs_2ai__usage.html).
- [x] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.
- [ ] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

Issue #171, opened 2025-06-05
✨ Pulse-Level Support


mnfarooqi:
🤖 *AI text below* 🤖

Lead: @burgholzer.

### What's the problem this feature will solve?

This issue focuses on extending the QDMI to include pulse level support, as discussed offline.
Exposing the pulse-level support will open up a range of new opportunities, such as calibration and enabling a pulse interface for advanced users. It will also enable the development of software that can optimise user programs at a lower level of abstraction.

Based on some preliminary discussions, the issue will be divided into the following sub-issues to facilitate subsequent discussions and PRs.

- #172 
- #173 
- #174 
- #175 
- #176 
- #177 

**Open questions**:

Do the supported pulse shapes on a device change from one site/channel to another, or are they global to the device?

FYI:
@burgholzer 
@ystade 
@echavarria-lrz 
@donsano33 
@amitQC

### Describe the solution you'd like

The solution will be implemented in a series of PRs.

## Scope and related work

QDMI #513 removes a coarse metadata declaration without removing provider-defined pulse-program submission. Keep byte-exact transport, metadata needed to construct/interpret a pulse program, and a standardized pulse programming interface distinct.

Existing transport: [IQM #200](https://github.com/iqm-finland/QDMI-on-IQM/pull/200). Concrete metadata requirements: [IQM #234](https://github.com/iqm-finland/QDMI-on-IQM/issues/234) and #523. Provider-specific calibration readiness and workflows can use documented extensions in v1. A general pulse interface remains a separate design topic; no implementation or release deadline for it is imposed here.

PR #513, opened 2026-08-24
💥 Remove unused pulse metadata

burgholzer:
🤖 *AI text below* 🤖

## Description

Remove `QDMI_DEVICE_PROPERTY_PULSESUPPORT` and `QDMI_Device_Pulse_Support_Level`, which have no corresponding pulse-programming interface. Keep provider-specific pulse programs possible without prescribing their design; requirements are tracked in #523.

This is the second metadata-cleanup layer: `v1.4` → #512 → #513. No dependency on payload capabilities, multi-program jobs, or replaceable drivers remains. The regular property-enum gap is closed; consumers must rebuild against matching 1.4 headers.

## Validation

- Release build: passed, including example and template targets.
- CTest: 81 passed; 18 expected read-only job tests skipped; no failures.
- `uvx prek run -a`: passed.
- Commit signed and verified. Fresh hosted CI is pending.

AI assistance: Codex separated the stack, preserved current develop tests, and ran local validation.

## Pulse submission boundary

Removing the coarse pulse metadata does not remove provider-defined pulse-program submission. Opaque payload transport, compiler-construction metadata, and a standardized pulse programming interface are separate concerns. Keep [IQM #200](https://github.com/iqm-finland/QDMI-on-IQM/pull/200) and [#234](https://github.com/iqm-finland/QDMI-on-IQM/issues/234) linked to #171/#523 without making transport wait for a universal pulse interface.

Before the interface change merges, link a working Core consumer and a demonstration with at least one existing provider, preferably both where applicable. Compatibility evidence suffices when a provider needs no changes. Development revisions may support these tests; published artifacts must use released dependencies. Core adoption: [#2233](https://github.com/munich-quantum-toolkit/core/pull/2233).

## Checklist

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://munich-quantum-software-stack.github.io/QDMI/latest/md_docs_2ai__usage.html).
- [ ] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -27,6 +27,15 @@ available. Removing the property shifts later regular device-property values
 down by one. Rebuild clients, drivers, and devices against matching QDMI 1.4
 headers; do not mix binaries built against different interim 1.4 revisions.
 
+### Pulse metadata
+
+The unused `QDMI_DEVICE_PROPERTY_PULSESUPPORT` property and
+`QDMI_Device_Pulse_Support_Level` type are removed. QDMI does not define a
+pulse-programming interface. Providers can expose pulse programming through
+custom program formats or a separate vendor interface. The removal shifts later
+regular device-property values down by one again; rebuild all clients, drivers,
+and devices against matching headers.
+
 ## [1.3.3]
 
 ### Retrieving existing jobs by ID
--- a/examples/device/src/cxx_device.cpp
+++ b/examples/device/src/cxx_device.cpp
@@ -802,10 +802,6 @@ int CXX_QDMI_device_session_query_device_property(
   ADD_LIST_PROPERTY(QDMI_DEVICE_PROPERTY_COUPLINGMAP, CXX_QDMI_Site,
                     DEVICE_COUPLING_MAP, prop, size, value, size_ret)
 
-  ADD_SINGLE_VALUE_PROPERTY(
-      QDMI_DEVICE_PROPERTY_PULSESUPPORT, QDMI_Device_Pulse_Support_Level,
-      QDMI_DEVICE_PULSE_SUPPORT_LEVEL_NONE, prop, size, value, size_ret)
-
   ADD_STRING_PROPERTY(QDMI_DEVICE_PROPERTY_LENGTHUNIT, "um", prop, size, value,
                       size_ret)
   ADD_SINGLE_VALUE_PROPERTY(QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR, double, 1.0,
--- a/include/qdmi/constants.h
+++ b/include/qdmi/constants.h
@@ -346,14 +346,6 @@ enum QDMI_DEVICE_PROPERTY_T {
    */
   QDMI_DEVICE_PROPERTY_COUPLINGMAP = 7,
   /**
-   * @brief @ref QDMI_Device_Pulse_Support_Level Whether the device supports
-   * pulse-level control.
-   * @details This property indicates the level of pulse-level control.
-   * If a device supports pulse-level control, it may provide additional
-   * functionality for pulse-level programming and execution.
-   */
-  QDMI_DEVICE_PROPERTY_PULSESUPPORT = 8,
-  /**
    * @brief `char*` (string) The length unit reported by the device.
    * @details The device implementation must report a known SI unit (e.g., "mm",
    * "um", or "nm") for this property. A client querying a length value must
@@ -361,7 +353,7 @@ enum QDMI_DEVICE_PROPERTY_T {
    * resulting value is then interpreted in the unit specified by this property.
    * @note If the device reports any length values, this property must be set.
    */
-  QDMI_DEVICE_PROPERTY_LENGTHUNIT = 9,
+  QDMI_DEVICE_PROPERTY_LENGTHUNIT = 8,
   /**
    * @brief `double` A scale factor for all length values.
    * @details The device implementation reports this scale factor. A client must
@@ -371,7 +363,7 @@ enum QDMI_DEVICE_PROPERTY_T {
    * @note If querying this property returns @ref QDMI_ERROR_NOTSUPPORTED, a
    * client should assume a default value of `1.0`.
    */
-  QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR = 10,
+  QDMI_DEVICE_PROPERTY_LENGTHSCALEFACTOR = 9,
   /**
    * @brief `char*` (string) The duration unit reported by the device.
    * @details The device implementation must report a known SI unit (e.g., "ms",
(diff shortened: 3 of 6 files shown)


mnfarooqi (commented):
@mortaza-lrz is working on a proposal for QDMI v2. However, if no other breaking changes are introduced in v1.4, keeping the enums intact makes sense. Removing them could create extra overhead for implementers who need to avoid breaking changes caused by shifting enum values.

burgholzer:
As you can see from this PR stack, QDMI v1.4 will likely contain a few more breaking changes/cleanups.
I'd rather remove aspects that have no credible use in QDMI v1.
If the enum removed here is still relevant for the pulse-level extension in QDMI v2, then it should become part of that module in my opinion.

mnfarooqi:
> As you can see from this PR stack, QDMI v1.4 will likely contain a few more breaking changes/cleanups. I'd rather remove aspects that have no credible use in QDMI v1. If the enum removed here is still relevant for the pulse-level extension in QDMI v2, then it should become part of that module in my opinion.

It's fine for the pulse enums it can be added in v2 when needed. :thumbsup: (The intended emoji appears to be disabled in the repository, or there is a GitHub glitch.)

I was just pointing out the issue of backward compatibility, and how inconvenient it would be for developers trying to develop software stacks that are compatible with all these different versions. Hopefully, from v2 onwards, backward compatibility will be in the list of priorities for QDMI.

burgholzer:
> Hopefully, from v2 onwards, backward compatibility will be in the list of priorities for QDMI.

Definitely. Already now, we are trying to be very mindful with breaking changes and only make them when they genuinely make sense and are driven by a particular use case that was not possible before the change.

I would not expect hosting entities to support multiple major versions of the interface in parallel.
I'd even hope that we can limit the amount of glue code necessary to support multiple minor versions. QDMI v1.x still has the contract that minor versions may introduce breaking changes. Maybe QDMI v2.x is the release where we can bump that to only major releases introducing truly breaking changes. We will have to see. The ecosystem is still developing and changing quickly. Which also means that the requirements for QDMI as an interface are changing quickly for the moment.

mnfarooqi:
Unfortunately, hosting entities cannot choose which version they support. It depends on the devices being hosted and the QDMI device version supported by their vendors. Some vendors would quickly adopt the newer version, while others would still only support the older version. For now, the only workaround is for hosting entities to have their own layer of abstraction to support multiple versions.

burgholzer:
Yeah. At MQSC, we are trying very hard to push and support vendors in a quick update/upgrade path, which would hopefully ease the situation in the future.

burgholzer:
@mnfarooqi this has now been adjusted to preserve the enumeration numbers, which should help with backwards-compatibility. At least, it will not require re-mapping of enum values.

examples/device/src/cxx_device.cpp
ystade: An exception for the remove property `...PROPERTY_PULSESUPPORT` must be added here, right?

ystade (changes requested):
LGTM besides the one request contained inline.

include/qdmi/constants.h
```diff
@@ -346,14 +346,7 @@ enum QDMI_DEVICE_PROPERTY_T {
...
   /// Value 8 is reserved for the removed calibration advisory; do not reuse it.
-  /**
-   * @brief @ref QDMI_Device_Pulse_Support_Level Whether the device supports
-   * pulse-level control.
-   * @details This property indicates the level of pulse-level control.
-   * If a device supports pulse-level control, it may provide additional
-   * functionality for pulse-level programming and execution.
-   */
-  QDMI_DEVICE_PROPERTY_PULSESUPPORT = 9,
+  /// Value 9 is reserved for the removed pulse metadata; do not reuse it.
```
denialhaag: Should this use `//` or `/** [...] */` instead?

UPGRADING.md
```diff
@@ -36,6 +36,17 @@ querying supported properties. Queries for value 8 return
 `QDMI_ERROR_NOTSUPPORTED`. Source code referring to the removed name must be
 updated when rebuilding against QDMI 1.4.
 
+### Pulse metadata
+
+The unused `QDMI_DEVICE_PROPERTY_PULSESUPPORT` property and
+`QDMI_Device_Pulse_Support_Level` type are removed. QDMI does not define a
+pulse-programming interface. Providers can expose pulse programming through
+custom program formats or a separate vendor interface. The removed property’s
```
denialhaag:
```suggestion
custom program formats or a separate vendor interface. The removed property's
```

UPGRADING.md
```diff
@@ -36,6 +36,17 @@ querying supported properties. Queries for value 8 return
...
+numeric value (9) remains reserved and returns `QDMI_ERROR_NOTSUPPORTED`. All
+remaining property values are unchanged. Existing binaries can continue querying
```
denialhaag: It might make sense to remove "All remaining property values are unchanged." because the release removes two device properties, right?

denialhaag (approved):
In addition to @ystade's comments, I only have three minor remarks (all of which are minor):

examples/device/src/cxx_device.cpp
burgholzer: See the other comment in #512. I personally would favor not adding these exceptions and simply have devices report NOTSUPPORTED.

include/qdmi/constants.h
burgholzer: Let's go with `/** [...] */`

UPGRADING.md
burgholzer: Yeah, right. We should either drop that or rephrase accordingly (e.g., by combining both fragments).

examples/device/src/cxx_device.cpp
ystade: As in #512. Let's stick with it.

UPGRADING.md
burgholzer: Adjusted in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/513/commits/af37fdaa6cc7f0d3acaf5f4a659caebcdc1411ff

UPGRADING.md
burgholzer: done in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/513/commits/af37fdaa6cc7f0d3acaf5f4a659caebcdc1411ff

include/qdmi/constants.h
burgholzer: changed in https://github.com/Munich-Quantum-Software-Stack/QDMI/pull/513/commits/af37fdaa6cc7f0d3acaf5f4a659caebcdc1411ff

denialhaag (approved):
:shipit:

ystade (approved):
The actual changes LPGTM (Look Pretty Good To Me ;) )

I am just wondering whether the return value NOTSUPPORTED should also be mentioned in the actual docstrings, not only in the UPGRADING.md file. What do you think?

Nevertheless, I am fine with incorporating these changes and hence approve.

